PR #220, opened 2026-03-05
Implement string escape sequences (#216)

### LINKED ISSUE #216: String escape sequences (\n, \t, etc.) not parsed in string literals
## Problem

The Vera grammar's `STRING_LIT` terminal regex does not correctly handle escape sequences like `\n`, `\t`, `\\\\`, etc. Writing `"hello\nworld"` causes a parse error instead of producing a string with an embedded newline.

## Root Cause

In `vera/grammar.lark`, the `STRING_LIT` regex uses Lark's string-based regex definition. The double-escaping interaction between Python string escaping and Lark's own regex compilation means `\\\\n` (intended to match the two-character sequence `\\n` in source) doesn't survive the escaping pipeline correctly.

## Current Workaround

Programs that need newlines use separate `IO.print` calls for each line, or avoid escape sequences entirely. The IO operation examples in #135 use this workaround.

## Expected Behavior

`"hello\nworld"` should parse to a string containing a literal newline character. Similarly for `\t` (tab), `\\\\` (backslash), `\"` (quote), and `\0` (null).

## Suggested Fix

Options:
1. Use a Lark `ESCAPED_STRING` import (built-in terminal) — simplest if it matches Vera's string semantics
2. Define a multi-part regex with explicit alternatives for each escape sequence
3. Use a raw string (`r"..."`) for the Lark regex definition to avoid Python-level escaping confusion

## Spec Reference

Section 4.3 (Literals) defines string literals as supporting standard escape sequences.

aallan:
## Summary
- Fix grammar regex (`STRING_LIT`) to accept single-backslash escape sequences — the original regex had too many backslashes and required `\\n` instead of `\n`
- Add `_decode_string_escapes()` to the transformer to decode all 7 spec-defined escapes: `\\`, `\"`, `\n`, `\t`, `\r`, `\0`, `\u{XXXX}`
- Add `_encode_string_escapes()` to the formatter to re-encode decoded characters back to escape sequences for round-trip correctness
- Add error code E009 for invalid escape sequences (e.g. `\x`)
- 14 new AST tests + 5 end-to-end codegen tests
- Update examples to use `\n` where applicable
- Bump version to 0.0.67

Fixes #216

## Test plan
- [x] 14 new tests in `test_ast.py` covering all 7 escape types + 4 error cases
- [x] 5 new tests in `test_codegen.py` for end-to-end execution with escape sequences
- [x] All 1,473 tests pass
- [x] mypy clean
- [x] All 18 examples pass `vera check` + `vera verify`
- [x] Spec, README, and SKILL.md code blocks all parse
- [x] Formatter idempotency tests pass (escape re-encoding round-trip)
- [x] Version sync verified


aallan:
## The three-layer escape problem

String escapes in Vera involved three layers of escaping:

1. **Lark grammar** — the regex `/\"([^\"\\]|\\.)*\"/` runs through Lark's own string processing, which converts `\"` to `"` but leaves `\\` pairs as-is
2. **Python regex engine** — interprets the processed pattern, where `\\.` means "backslash followed by any char"
3. **Transformer** — decodes Vera escape sequences (`\n` to newline) into Python string values

The original bug had too many backslashes in layer 1, causing the regex to require *double* backslashes in Vera source. The fix reduced the escaping to match what the spec defines.

## Round-trip invariant

The formatter must be the exact inverse of the transformer for string literals: `encode(decode(s)) == s`. This is verified by the idempotency tests (`format(format(x)) == format(x)`), which caught the missing re-encoding immediately.

Issue #522, opened 2026-04-23
IO.print output lost when program traps (stdout fully buffered)


aallan:
**Summary:** Output written via `IO.print` before a runtime trap is silently discarded. The captured stdout buffer is only returned to the CLI after successful execution; on trap the buffer is dropped as the exception unwinds.

## Reproduction

```vera
public fn main(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
  IO.print("line 1 before crash\n");
  IO.print("line 2 before crash\n");
  IO.print("line 3 before crash\n");
  IO.print("line 4 before crash\n");
  let @Nat = 42 / 0;
  ()
}
```

```console
$ vera run /tmp/buffer_test.vera
Error: Runtime contract violation: error while executing at wasm backtrace:
    0:     0x62 - <unknown>!main

Caused by:
    wasm trap: integer divide by zero
```

Expected: the four `line N before crash` lines precede the trap message. Actual: all four are lost.

## Root cause

The `IO.print` host import in `vera/codegen/api.py` (`host_print`, ~line 450) appends the decoded UTF-8 to a Python `io.StringIO` named `output_buf` — it never writes to `sys.stdout` directly. The buffer is only surfaced to the CLI after the WASM call completes successfully:

```python
# vera/codegen/api.py:2442-2475 (abridged)
try:
    raw_result = func(store, *call_args)
except _VeraExit as exit_exc:
    return ExecuteResult(stdout=output_buf.getvalue(), ...)   # exit(): OK
except Exception as exc:
    # ...
    if exc_name in ("Trap", "WasmtimeError") and last_violation:
        raise RuntimeError(last_violation[0]) from exc   # <-- buffer discarded
    raise                                                # <-- buffer discarded
```

On trap, the `raise` path unwinds without doing anything with `output_buf`. The CLI (`vera/cli.py:605`) only prints `exec_result.stdout` *after* `execute()` returns normally. So on any trap, every prior `IO.print` byte is thrown away.

This is not the wasmtime runtime buffering anything — wasmtime-py calls `host_print` synchronously for each `IO.print` at the Vera source site; the buffering is entirely in our Python shim.

## Impact

- Debugging by inserting `IO.print` calls before a suspected crash site doesn't work — the prints never reach the terminal.
- Conway's Game of Life session lost ~30 minutes to this: the agent inserted instrumentation prints, saw no output, concluded its changes weren't being exercised, and went looking for the wrong bug.
[...]

Issue #516, opened 2026-04-23
Runtime traps need Vera-native diagnostics, not raw wasmtime stack traces


aallan:
## Summary

Runtime traps from the WASM runtime bubble up as raw wasmtime stack traces with hex offsets, not as Vera diagnostics. Compile-time errors have carefully crafted "Fix:" suggestions; runtime traps don't. An agent hitting a GC bug / integer overflow / bounds check trap has no actionable information.

## Example

From the Game of Life agent run that surfaced [#515](https://github.com/aallan/vera/issues/515):

```
Error: Runtime contract violation: error while executing at wasm backtrace:
    0:    0x33e - gc_collect
    1:    0x1b9 - alloc
    2:    0x55f - cell_at
    ...

Caused by:
    0: memory fault at wasm address 0x180000 in linear memory of size 0x180000
    1: wasm trap: out of bounds memory access
```

Three problems in this one output:

1. **"Runtime contract violation" is often wrong**: `vera/cli.py:654` blanket-labels every `Trap`/`WasmtimeError` as a contract violation:

    ```python
    if exc_name in ("Trap", "WasmtimeError"):
        msg = f"Runtime contract violation: {exc}"
    ```

    But traps include out-of-bounds memory access (GC bug or array OOB), integer overflow (INT64 saturation), unreachable (compiler asserts), and stack overflow — none of which are contract violations. Agents and humans reading the first line dismiss real compiler bugs as "my contract is wrong."

2. **Stack offsets are hex WASM addresses**: `0x33e` tells nobody anything. Even with symbol names (`gc_collect`, `alloc`), there's no source-line mapping — runtime traps from user functions like `cell_at` don't tell the user which line of `cell_at` faulted.

3. **No actionable guidance**: Compile-time errors are a paragraph — what went wrong, rationale, fix, spec reference. Runtime traps are a stack trace. Contrast `E130`:

    ```
    Unresolved slot reference. @Int.3 has no matching binding.
    Fix: ...
    ```

    with the trap above. The former helps; the latter requires WASM internals to interpret.

## Proposed scope

Staged improvements, each an independent PR:

### Stage 1: Categorise the trap reason (small, high impact)

Parse the `wasm trap: ...` tail of the error and map to distinct human categories:

| Trap reason | User message |
|---|---|
[...]

PR #532, opened 2026-04-27
docs(site): fix three mobile overflow bugs (hero image, bench button, runs code blocks)

aallan:
## Summary

Three mobile overflow bugs in `docs/index.html` reported on iOS Safari at the iPhone width:

1. **Hero meerkat image** ran off the right edge of the viewport.
2. **VeraBench `vera-bench` GitHub button** was stretched to a full-width chunky pill instead of sitting inline next to its caption.
3. **Shell-command code blocks** in the "Runs Everywhere" section ran past the right edge of the viewport, hiding text like `--fn factorial -- 10`.

Each had a different root cause; each is fixed with a small CSS change.

### 1. Hero image overflow

`.hero-image` had `max-width: 640px` but no `width: 100%`. The global `img { max-width: 100% }` rule was overridden by the more-specific class selector, so on a 375px viewport the image kept a 640px target and overflowed.

```diff
-.hero-image{max-width:640px;margin:0 auto;border-radius:10px}
+.hero-image{width:100%;max-width:640px;margin:0 auto;border-radius:10px}
```

### 2. VeraBench button stretched

The mobile media query had `.btn { width: 100%; max-width: 280px }` to stretch hero CTA buttons (which stack vertically and benefit from a full-width target). That rule was matching every `.btn` on the page, including the isolated `vera-bench` GitHub button under the bench chart, which has no need to stretch.

Scoped to the CTA bar only:

```diff
-.btn{width:100%;max-width:280px;justify-content:center}
-.cta-bar{flex-direction:column;align-items:center}
+.cta-bar{flex-direction:column;align-items:center}
+.cta-bar .btn{width:100%;max-width:280px;justify-content:center}
```

### 3. Shell code blocks overflow

`<pre>` defaults to `white-space: pre` (no wrapping). Lines like `vera run examples/factorial.vera --fn factorial -- 10` are wider than a 327px content column. `overflow-x: auto` produces a horizontal scrollbar inside the box, but iOS has no good gesture for scrolling inside a narrow card.

Added a mobile-only rule that lets shell-command blocks wrap:

```diff
+.code-block pre{white-space:pre-wrap;word-break:break-word}
```

Vera sample blocks (`.sample .code`) intentionally keep `pre` so syntax indentation isn't disrupted at narrow widths — only `.code-block` (used by the runs/install/agents sections, which all hold shell commands) gets the wrap treatment.

## Verification

At 375x812 viewport:

| Element | Before | After |
|---|---|---|
| `.hero-image` width | 640px (overflowed) | 327px (fits, vw=375) |
| `.bench-body .btn` width | 280px (stretched) | 144px (inline-sized) |
| `.cta-bar .btn` width | 280px | 280px (still stretched, by design) |
| `.runs .code-block` overflow | yes | no (`scrollWidth === clientWidth`) |

## Test plan

- [x] DOM-level checks at 375x812 confirm all three fixes
- [x] Mobile screenshot shows hero meerkat + wordmark both fully visible
- [x] Hero CTA stack still stretches as designed at 375px
- [ ] CI green after push


aallan:
Closing — folded into the v0.0.120 release PR.

The single commit (`7911ad8 docs(site): fix three mobile overflow bugs`) was cherry-picked onto the `bug-campaign-trap-diagnostics` branch alongside the #522 + #516 Stage 1 work. Authorship preserved; the CSS-only diff is unchanged. CHANGELOG entry under v0.0.120 `### Website` credits this PR by number.

Rationale for folding rather than separate-merge:

- Single-file (`docs/index.html`), CSS-only — no overlap with the trap-handling code that the v0.0.120 PR also touches in `docs/index.html`'s version-bump line.
- v0.0.120 was already going to update `docs/index.html` for the version bump (`v0.0.119` → `v0.0.120` in the `<a href=...releases/tag/v0.0.119>` block), so combining avoids two near-simultaneous edits to the same file.
- Mobile overflow is a release-quality fix; better to ship it inside the version it's first present in than as an immediate follow-up `v0.0.120.1`.

The v0.0.120 PR is now open as #533. No action needed here.

Issue #348, opened 2026-03-25
GC worklist overflow for deeply nested object graphs


aallan:
The mark-sweep GC in `vera/codegen/assembly.py` uses a fixed-size worklist of 1024 entries (4096 bytes / 4 bytes per entry). If a deeply nested or wide object graph exceeds this capacity, reachable objects beyond the worklist limit will not be marked and will be incorrectly freed during the sweep phase.

## Impact

Programs with deeply nested ADT structures (e.g., a large Json tree, deeply nested Option chains, or long linked lists) could trigger premature collection of reachable objects, causing use-after-free bugs.

## Current mitigation

The 1024-entry limit is sufficient for typical programs. The risk increases with recursive ADT types like Json (`JArray(Array<Json>)` containing nested `JObject(Map<String, Json>)`).

## Possible fix

- Dynamic worklist growth (realloc when full)
- Iterative deepening (re-scan from roots if worklist overflows)
- Increase the fixed size (simple but doesn't eliminate the risk)

**Affects:** All heap-allocated types. Higher risk with recursive ADTs like Json.

PR #542, opened 2026-04-27
v0.0.122: bound conservative GC scan against $heap_ptr (#515)

### LINKED ISSUE #515: GC collect itself faults with out-of-bounds memory access under sustained allocation pressure
## Summary

`$gc_collect` itself faults with out-of-bounds memory access when collecting under sustained allocation pressure. The collector is the top frame of the crashing stack, not the user code that triggered it. A 40×20 Game of Life grid running 200 generations reliably reproduces this on v0.0.119.

## Reproducer

Attach `life.vera` (40×20 grid, 200 generations, builds each generation via tail-recursive `array_append` over an immutable `Array<Array<Bool>>`, with each cell lookup going through an `Option<Nat>` from `int_to_nat`). ~14,400 small allocations per generation.

```
vera run life.vera
```

Consistently crashes at some generation during `next_row_rec`:

```
    0:    0x33e - gc_collect
    1:    0x1b9 - alloc
    2:    0x55f - cell_at
    3:    0x662 - count_neighbors
    4:    0x6c3 - next_cell
    5:    0x9ce - next_row_rec
    ... (deep recursion of next_row_rec)

Caused by:
    0: memory fault at wasm address 0x180000 in linear memory of size 0x180000
    1: wasm trap: out of bounds memory access
```

The faulting address `0x180000` equals the current memory size — the GC is walking one byte past the end of linear memory.

## Why this isn't user OOM

- 40×20 = 800 cells. Grid state is trivially small.
- Same program with 10×10×10 runs to completion.
- The fault is inside `$gc_collect`, not inside `$alloc` failing to find a slot.
- The memory is exactly at the page boundary (`0x180000` = 24 × 64 KB).

The collector itself shouldn't fault regardless of how much pressure the mutator puts on it. A sweep that walks the heap needs to stop at `$heap_ptr`, not at the linear-memory bound — or memory-grow-on-exhaust if it's trying to mark beyond the current allocation.

## Related

- [#487](https://github.com/aallan/vera/issues/487) — `$alloc` grows memory by only 1 page; single large requests trap. That's about the mutator's allocator path; this is about the collector's sweep path. They may share a root cause (memory growth policy during GC) but have distinct symptoms.
- [#484](https://github.com/aallan/vera/issues/484) fixed the sweeper's 16-bit size-field truncation. This is a separate crash — the sweep loop's bound check is wrong in a different way.

## Initial investigation pointer

The fault is inside `$gc_collect` at WASM offset `0x33e`. Inspect `_emit_gc_collect` in `vera/codegen/assembly.py` — specifically the Phase 2 mark-seed loop and the Phase 3 sweep loop — and look for any load/store that uses a pointer without an explicit `< $heap_ptr` bound check, or any `memory.grow` call missing its return-value check.

Discovered by an agent writing Conway's Game of Life in Vera. Same agent whose earlier feedback drove the Stage 11 primitives push.

aallan:
Closes #515.

## Summary

`$gc_collect` was faulting under sustained allocation pressure with `gc_collect` itself at the top of the stack — the collector, not the program, was the crashing frame. Symptom: `memory fault at wasm address 0x... in linear memory of size 0x...`. Reliably reproduced with a 470-line Conway implementation crashing at generation 56.

## Root cause

The Phase 2 worklist-seeding code in `_emit_gc_collect` accepts a shadow-stack value as a heap pointer if it satisfies three guards:
- in heap range (`val >= gc_heap_start + 4`)
- properly aligned (`(val - gc_heap_start) % 8 == 4`)
- below `$heap_ptr`

None of those guards prove the word at `val - 4` is actually an object header. A non-pointer i32 in payload data (e.g. a bit-packed `Nat` row in Conway-style code, a hash, anything else with bits in heap range) can satisfy all three; Phase 2b then reads garbage as `obj_size = header >> 1`, sets the mark bit (corrupting a random heap word!), and walks `obj_ptr + 0..obj_size` past `$heap_ptr` and past the linear-memory boundary.

This is the classic conservative-GC false-positive failure mode.

## Fix

Two layers of defence in `_emit_gc_collect` (`vera/codegen/assembly.py`):

- **Layer 2 (early skip)** — before either marking or scanning a worklist entry, compute `obj_size` and verify `obj_ptr + obj_size <= heap_ptr`. If not, skip the entry entirely (no mark store, no scan loop). Catches the bug at the cheapest possible point and prevents the mark-bit corruption that would otherwise persist across the cycle.
- **Layer 1 (per-iter)** — inside the conservative scan loop, also check `obj_ptr + scan_ptr + 4 <= heap_ptr` before each `i32.load`. Negligible cost relative to the load it guards, and protects any future caller that reaches this loop without the Layer-2 check.

## Tests

- Structural regression test: `tests/test_codegen.py::TestGarbageCollection::test_gc_collect_bounds_check_against_heap_ptr` asserts both bound checks survive in the emitted WAT. Behavioural reproducers for #515 are heavily layout-sensitive (string-pool offsets, allocation order), so a structural assertion is the durable regression guard.
- 3,569 tests pass; mypy clean; 81 conformance programs pass; 33 examples pass.

## Verification

- 470-line Conway implementation now runs cleanly through every generation (was crashing at gen 56 pre-fix).

## Test plan
- [x] `pytest tests/ -q` — 3,555 passed, 14 skipped
- [x] `mypy vera/` — clean
- [x] `python scripts/check_conformance.py` — all 81 pass
- [x] `python scripts/check_examples.py` — all 33 pass
- [x] `python scripts/check_version_sync.py` — version 0.0.122 consistent
- [x] All pre-commit hooks pass

Co-Authored-By: Claude <noreply@anthropic.invalid>

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a GC crash occurring under sustained allocation pressure by adding defensive heap-bounds checks during conservative GC scanning.

* **Tests**
  * Added a regression test to verify the conservative-scan bounds and prevent regressions.

* **Chores**
  * Bumped release version to v0.0.122 and updated package metadata.

* **Documentation**
  * Updated changelog, roadmap, README, testing metrics and module GC docs to reflect the fix and new release.


aallan:
@coderabbitai thanks. Three of four findings actioned in 75682a1:

- **ROADMAP #348 worklist size** — fixed: now reads "16 KiB / 4,096-entry cap (`gc_worklist_size = 16384` in `vera/codegen/assembly.py`)". You were right that the existing "1024-entry cap" was stale.
- **HISTORY footer** — fixed: "121 tagged releases" → "122 tagged releases" (the v0.0.122 row was already in the Stage 9 release table; the footer hadn't been bumped).
- **Test assertions on bound-check opcodes** — fixed: the test now asserts the exact opcode pattern adjacent to each `Layer N (issue #515)` marker, not just the marker text. A refactor that left the comment in place but deleted the underlying check fails the test now. Marker comments stay as discoverability anchors; opcodes are the contract.

For the fourth (outside-diff in `vera/codegen/assembly.py`, "load and validate the actual candidate header at `potential_header_ptr = $obj_ptr - 4`"): not actioning because the code already does this. Lines 674-679 load `$header` from `$obj_ptr - 4` directly:

```wat
local.get $obj_ptr
i32.const 4
i32.sub
i32.load
local.set $header
```

`$header` IS the value at `$obj_ptr - 4` — it's the `potential_header_ptr` value the suggestion asks for, just under a different name. `$obj_size` is decoded from it (line 681-684), the bound check `obj_ptr + obj_size <= heap_ptr` runs against it (lines 695-702), the already-marked check runs against it (lines 704-708), and the mark store writes back to `$obj_ptr - 4` with the validated value (lines 711-717). The header-tag invariant in this codebase is `(size << 1) | mark` with no separate "is-this-a-header" tag bit — the bound check on `obj_size` is the only validation possible without introducing a new tag scheme, and that's exactly what Layer 2 does. Happy to rename `$header` → `$candidate_header` if that would make the intent clearer, but the validation is structurally complete.

Also pushed `uv.lock` regenerated for the version bump (lint job was failing on `uv lock --check`).

Issue #608, opened 2026-05-07
Doc: terminal-vs-browser IO seam in 'write once run anywhere' framing


aallan:
## Summary

Vera's "write once, run anywhere" framing has a real seam at the IO boundary that's not currently surfaced in user-facing documentation. A program written for terminal-target ergonomics (using `IO.sleep` for animation timing and ANSI escape codes for cursor control) compiles cleanly to `--target browser` but doesn't *run* meaningfully there: `IO.sleep` busy-waits and freezes the tab, and ANSI escapes render as literal text in the DOM. The right shape for the browser is "Vera as pure simulation core, JS as timing+rendering driver" — but this isn't documented anywhere, and the runtime gaps that would make the JS-driver pattern ergonomic ([#603](https://github.com/aallan/vera/issues/603)) aren't filled in either.

## Concrete manifestation

An agent writing Conway's Life on current main:

1. Wrote a perfectly reasonable terminal-target program (`IO.sleep(100)` between frames + ANSI `\u{1B}[2J\u{1B}[H` clear+home).
2. Compiled it cleanly with `vera compile --target browser`.
3. Tried to load the bundle and observed: tab frozen for 30 seconds, output appearing as raw escape codes in DOM rather than animated grid.
4. Wrote a separate browser-shaped variant: pure simulation core in Vera, JS driver using `requestAnimationFrame` and `<canvas>`.
5. Hit [#603](https://github.com/aallan/vera/issues/603) (no string-marshalling helpers exposed), worked around with a "compute-everything-upfront, drain stdout once" pattern.

This is a real adoption surface: the message "Vera is write-once-run-anywhere" sets up an expectation that the same source runs identically on both targets. The seam at `IO` breaks that expectation in a way that isn't a bug — it's a runtime-shape mismatch that no language can fully hide — but the documentation should set the expectation correctly.

### Option 1 (recommended): embrace the seam, document it explicitly

- Add a "Browser target: limitations and patterns" section to SKILL.md and the spec
- Explicitly call out: terminal animations using `IO.sleep` + ANSI escapes don't translate; the browser runtime expects "Vera = pure simulation core, JS = timing/rendering" and that the IO surface for the browser is `IO.print` (writes to the page) only
- Provide a worked example (Conway's Life is a candidate) showing the JS-driver pattern
- Pair with [#603](https://github.com/aallan/vera/issues/603) to make the marshalling ergonomic

### Option 2: bridge the seam in the runtime

[...]

Issue #604, opened 2026-05-07
Five prelude combinators silently skipped from every WASM compile (option_map / option_and_then / result_map + two _unwrap_or variants)


aallan:
# Investigation: Five prelude combinators are silently skipped from every WASM compile

**Component:** WASM backend (codegen) + standard prelude
**Severity:** Medium — silent feature gap, not a crash
**Vera version:** `0.0.138` (git `20f3ccb`)
**Found via:** `vera compile --target browser` on any program

## Note on framing

In an earlier conversation I dismissed the warning about `result_map` being
"skipped" as irrelevant prelude noise that didn't affect the program I was
compiling. That dismissal was technically correct for that program but was
the wrong inference about the underlying state of the compiler. On a closer
look, **five** prelude functions are skipped from *every* WASM compile, and
the reasons split into two distinct backend gaps. This is worth a separate
look — both as a real feature gap for anyone writing browser-target Vera that
does monadic Option/Result work, and as a quality signal about how much of
the standard prelude actually round-trips through the WASM backend.

## Reproducer

Any non-trivial Vera program will surface these. Even the minimum reproducer
from the string-interpolation report (`hello\n` printer, 14 lines) emits
five `[E602]` warnings during browser compile:

```
$ vera compile --target browser hello.vera -o /tmp/out/
  Function 'option_unwrap_or' has unsupported parameter type — skipped.
warning: [E602] Warning at hello.vera, line 20, column 9:

  Function 'option_map' body contains unsupported expressions — skipped.

  The WASM backend does not yet support all Vera expression types. This
  function will not appear in the compiled output.

  Function 'option_and_then' body contains unsupported expressions — skipped.
  Function 'result_unwrap_or' has unsupported parameter type — skipped.
  Function 'result_map' body contains unsupported expressions — skipped.
```

(The line/column numbers point into the user's file, not the prelude, which
made the warnings feel like noise about user code rather than about the
prelude itself. That framing also probably contributed to my initial dismissal.)

## The five affected functions

Definitions extracted verbatim from `vera/prelude.py`:

```vera
private forall<T> fn option_unwrap_or(@Option<T>, @T -> @T)
  requires(true) ensures(true) effects(pure)
{
  match @Option<T>.0 { None -> @T.0, Some(@T) -> @T.0 }
}

private forall<A, B> fn option_map(@Option<A>, @OptionMapFn<A, B> -> @Option<B>)
  requires(true) ensures(true) effects(pure)
{
[...]

Issue #602, opened 2026-05-07
String-returning function call in string interpolation produces invalid WASM (i64/i32 mismatch)


aallan:
## Summary

`vera check` accepts a program that interpolates a `String`-returning function call into a string literal, but `vera compile` / `vera run` emits invalid WASM:

```
Error: Unhandled WASM trap: failed to compile: wasm[0]::function[N]::main
Caused by:
    0: WebAssembly translation error
    1: Invalid input WebAssembly code at offset NNN: type mismatch: expected i64, found i32
```

Type-check is silent; the failure surfaces only at the wasmtime validation step, and the error points at the WASM offset rather than the offending source line.

## Minimum reproducer

```vera
effect IO { op print(String -> Unit); }

private fn make(@Unit -> @String)
  requires(true) ensures(true) effects(pure)
{
  "hello"
}

public fn main(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
  IO.print("\(make(()))\n")
}
```

`vera check`: OK.
`vera run`: `type mismatch: expected i64, found i32` at WASM offset 911.

## What works (i.e. narrowing)

```vera
-- WORKS: string_concat instead of interpolation
IO.print(string_concat(make(()), "\n"))

-- WORKS: interpolating a slot ref of String type
let @String = "hello";
IO.print("\(@String.0)\n")

-- WORKS: interpolating an Int-returning call
private fn make_int(@Unit -> @Int) ... { 42 }
IO.print("\(make_int(()))\n")
```

So the trigger is specifically:
1. Expression is a function call (not a slot ref or literal)
2. **Return type is `String` (i32_pair)**
3. Used as an interpolation segment inside a string literal

The most likely root cause: the interpolated-string lowering treats the expression as i64 (the `to_string` path for numerics) without first checking whether it's already a String — for slot refs the type info propagates but for function calls it doesn't, so an i32_pair gets fed into an i64-expecting `to_string` site.

## Discovery context

Surfaced by an agent writing a browser variant of Conway's Life. The agent had the workaround in mind (use `string_concat`) within minutes, but the path from "vera check passes" to "the WASM offset 4179 is the i64-vs-i32 bug" was opaque — they had to bisect the program by removing constructs until the trap stopped firing.

For an LLM agent the friction is real:
- The error is at run-time, not compile-time, so the standard "fix the diagnostic" loop doesn't engage
- The error message points at WASM bytes, not Vera source
[...]

Issue #597, opened 2026-05-07
Walker-completeness audit: every isinstance(expr, ast.X) dispatch chain enumerates all Expr subclasses


aallan:
## Summary

Audit every `isinstance(expr, ast.X)` dispatch chain in the codebase and verify each one explicitly handles or explicitly skips every `Expr` subclass. Build a per-walker checklist of every `Expr` in `vera/ast.py` and document the disposition in code comments.

Discovered the need while diagnosing [#588](https://github.com/aallan/vera/issues/588): `_walk_free_vars` in `vera/wasm/closures.py` was missing **eight** `Expr` subclass branches (`IndexExpr`, `ArrayLit`, `InterpolatedString`, `HandleExpr`, `AssertExpr`, `AssumeExpr`, `ForallExpr`/`ExistsExpr`, `ModuleCall`). When the walker hit one of those expressions inside a closure body, it silently fell through, missed any captured slot references inside the sub-expressions, and the closure-lift produced a broken WAT module that failed validation or trapped at runtime.

The fix was mechanical (~50 lines), but the bug class is *systemic*: the original `_walk_free_vars` walked the obvious shapes (`BinaryExpr`, `FnCall`, `MatchExpr`) and missed the long tail of less-common expression types. We don't know how many other walkers in the codebase have the same shape of incompleteness.

## Why now

The stabilisation discussion identified this as an unknown: **was `_walk_free_vars` the worst offender, or is it representative?** We have at least these AST-walker-shaped functions in the codebase:

- `vera/wasm/closures.py::_walk_free_vars` — recently audited, was the trigger
- `vera/wasm/context.py::_is_void_expr` — handles `QualifiedCall`, `UnitLit`, `FnCall`, `AssertExpr`/`AssumeExpr`, `MatchExpr`, `IfExpr`, `Block`, `HandleExpr`. Missing IndexExpr / ArrayLit / etc.? Unknown.
- `vera/wasm/inference.py::_infer_expr_wasm_type` — large dispatch chain
- `vera/wasm/inference.py::_infer_vera_type` — parallel dispatch chain
- `vera/codegen/...` translate_expr family
- `vera/checker.py` synth/check dispatch
- `vera/verifier.py` SMT encoding dispatch
- `vera/ast.py::format_expr` — pretty-printer dispatch (less critical — format failures are visible)

Each of these is a candidate for the same bug class. Until audited, "the walkers are complete" is unverified.

## Proposed approach

For each walker:

[...]

PR #611, opened 2026-05-07
ROADMAP: add browser-target reliability (#609 + #610) to stabilisation tier

### LINKED ISSUE #603: Browser runtime: export string-marshalling helpers so JS can pass String args to Vera functions
## Summary

`vera/browser/runtime.mjs` exports 12 surface APIs (`init`, `call`, `getStdout`, `getState`, etc.) but does not export the internal string-marshalling helpers (`allocString`, `readString`, `allocArrayOfStrings`). This blocks the natural "Vera as pure simulation core, JS as driver" pattern in the browser, because there's no way to pass a `String` argument from JS into a Vera function call.

## Discovery context

Surfaced by an agent writing a browser variant of Conway's Life. The agent identified the natural decomposition:

- Vera holds the pure simulation: `step(grid) -> grid` is `effects(pure)` and bit-identical to the terminal version
- JS holds the timing and rendering: `requestAnimationFrame` driving a `<canvas>`, calling `step` once per frame

Per-frame call shape: `call("step_packed", currentStateString)`. But `call(fnName, ...args)` only accepts numeric primitives — there's no marshalling for `String` (or `Array<T>`) arguments. The internal helpers exist (`allocString` at line 101 of `runtime.mjs`, `readString` at line 55, `allocArrayOfStrings` at line 181) but they're file-private.

## Workaround the agent used

Compute all 300 generations upfront in Vera, emit each as a packed string of `0`/`1` chars to stdout, then `getStdout()` once and split on newlines:

```vera
-- Vera side: recursive emit_frames called from main, prints all 300 frames
private fn emit_frames(@Array<Array<Bool>>, @Nat -> @Unit) ...

-- main runs the whole simulation up-front
public fn main(@Unit -> @Unit) ... {
  let @Array<Array<Bool>> = make_grid(80, 22);
  IO.print(string_concat(pack_grid(@Array<Array<Bool>>.0), "\n"));
  emit_frames(step(@Array<Array<Bool>>.0), 300)
}
```

```js
// JS side: read all frames at once, animate via rAF
await init('module.wasm');
call('main');
const frames = getStdout().split('\n').filter(Boolean);
let i = 0;
function tick() {
  if (i < frames.length) { renderFrame(frames[i++]); requestAnimationFrame(tick); }
}
tick();
```

This works for fixed-length simulations (300 × 1760 chars ≈ 528 KB, computes in ~100 ms) but rules out:
- **Streaming / interactive simulations** where the user can perturb the state between frames
- **Long-running or unbounded simulations** that don't fit in memory
- **Per-frame UI feedback** where Vera needs to read input back from the page

## Proposed fix

Export the existing internal helpers from `vera/browser/runtime.mjs`:

```js
export function allocString(str) { ... }       // already exists at line 101
export function readString(ptr, len) { ... }    // already exists at line 55
export function allocArrayOfStrings(strs) {...} // already exists at line 181
```

Plus a thin wrapper on `call` that auto-marshals JS strings to `(ptr, len)` pairs:

```js
export function callWithStrings(fnName, ...args) {
  const flat = [];
  for (const a of args) {
    if (typeof a === 'string') {
      const [p, l] = allocString(a);
      flat.push(p, l);
    } else {
      flat.push(a);
    }
  }
  return call(fnName, ...flat);
}
```

Plus a corresponding return-value reader for `String`-returning functions.

## Why this matters for adoption

The terminal-vs-browser split exposed by Conway's Life is a real boundary in Vera's "write once, run anywhere" model — terminal animations rely on ANSI escapes and `IO.sleep` (main-thread blocking) that the browser runtime doesn't translate. The right pattern for the browser is "Vera computes pure state transitions, JS drives timing and IO," and this gap is the missing piece for that pattern to be ergonomic.

Documented in [SKILL.md](https://github.com/aallan/vera/blob/main/SKILL.md) browser-runtime section as a follow-up to make the split explicit once the helpers are exported.

## Acceptance

- `runtime.mjs` exports `allocString`, `readString`, and a `callWithStrings` (or equivalent auto-marshalling) wrapper
- Round-trip example test: a Vera function `fn(@String -> @String)` is callable from JS with a JS string and returns a JS string
- Browser-runtime SKILL section documents the JS-driver pattern with a worked example (Conway's Life is a candidate)

### LINKED ISSUE #609: Browser runtime: implement IO.sleep via JSPI (or Asyncify fallback) so animations don't freeze the tab
## Summary

`IO.sleep` in the browser runtime is currently a busy-wait that pegs the main thread for the entire duration. This makes any animation or simulation that uses `IO.sleep(N)` between frames freeze the tab — the canonical terminal-style render-sleep-step loop is fundamentally incompatible with a browser. The proposed fix doesn't require a language change: implement `IO.sleep` against the [WebAssembly JavaScript Promise Integration (JSPI)](https://github.com/WebAssembly/js-promise-integration) proposal, with [Asyncify](https://emscripten.org/docs/porting/asyncify.html) as the fallback for older browsers.

## Discovery context

An agent writing a browser variant of Conway's Life on current main reported this as the *core* obstacle to "write once, run anywhere": they had to fork their working terminal program (`life.vera`) into a browser-shaped variant (`life_web.vera`) with packed-string emission and a JavaScript canvas driver, purely because the terminal version's `IO.sleep(100)` between frames couldn't yield to the browser event loop.

Their framing — quoted because it's exactly right — was that this is "two things masquerading as one: timing and rendering. `IO.sleep` busy-waits because WASM can't yield to the JS event loop from synchronous code. ANSI escapes don't render because the browser runtime treats `IO.print` as text-to-DOM rather than text-to-terminal. Both meant `life.vera` couldn't run unchanged — I had to write a different program that computed the same thing differently."

This issue covers the timing half. The rendering half is tracked separately as a sibling architectural issue: ANSI subset interpretation in the browser runtime.

## Proposed approach

**Primary**: implement `IO.sleep` in `vera/browser/runtime.mjs` using JSPI:

1. The Vera-emitted WASM declares `vera.sleep(ms: i64)` as an imported function.
2. The runtime imports it as a JSPI-suspending function: `WebAssembly.promising` wraps a JS function that returns a `Promise` resolved by `setTimeout(resolve, ms)`.
3. Vera-side code calling `IO.sleep(100)` then suspends the WASM call, yields to the event loop, and resumes after 100ms — without blocking the main thread.

**Fallback**: if JSPI's browser support is too thin (it shipped in Chrome 137 and Firefox 134; Safari has no current implementation), provide an Asyncify-built variant of the same module that achieves the same suspension via stack-saving transformation. Asyncify has a runtime cost (~50% slowdown for code that uses the suspension paths), but it's a known-working drop-in.

The runtime should pick the right implementation at `init` time based on feature detection.

## Why this is high-leverage

- **It closes the timing half of the terminal-vs-browser seam without language changes.** The same `life.vera` source — `IO.sleep` between frames, recursive `run_loop` — would Just Work in both targets. No `life_web.vera` fork, no JS driver.
- **Pairs naturally with ANSI interpretation** (sibling issue): with both fixes, a typical terminal Vera program (animation via `IO.sleep` + screen control via ANSI escapes) renders identically on both targets.
- **No bundle-size cost in the JSPI path** — it's a thin runtime change. Asyncify adds bytes but only when JSPI is unavailable.

## Acceptance

- A program calling `IO.sleep(100)` in a recursive loop renders animated output in the browser without freezing the tab.
- The same source (`life.vera`) runs on both `vera run` (terminal) and `vera compile --target browser` (browser) with identical output.
- A test in `tests/test_browser.py` verifies that `IO.sleep` actually yields rather than busy-waiting (e.g., by measuring that other event-loop callbacks fire during a 500ms sleep).

## Related

- [#608](https://github.com/aallan/vera/issues/608) — the framing issue ("terminal-vs-browser IO seam")
- [#603](https://github.com/aallan/vera/issues/603) — runtime helper exports (separate gap, also browser-target)
- [#604](https://github.com/aallan/vera/issues/604) — prelude combinators skipped on browser target
- The agent's full design memo is posted as a comment on the umbrella issue: [#608 design memo](https://github.com/aallan/vera/issues/608#issuecomment-4399625656).  Key paragraph: *"JSPI-driven IO.sleep plus a minimal ANSI interpreter would do most of the heavy lifting on their own."*

### LINKED ISSUE #610: Browser runtime: interpret a minimal ANSI escape subset so terminal-style Vera programs render unchanged
## Summary

The browser runtime currently treats `IO.print` as text-to-DOM (writes go to a `<pre>` element appended to the page) without interpreting any ANSI escape sequences. Terminal-style Vera programs that use ANSI escapes for cursor positioning, screen clearing, or coloring — completely standard for animated terminal output — render their escape codes as literal text in the DOM, defeating the animation.

Bounded-scope fix: have the browser runtime interpret a small ANSI subset (cursor positioning, screen clearing, line erase, basic colors) and translate them into DOM mutations on a target `<pre>` element. Estimated ~200 lines of JS. Pairs with [the JSPI-driven `IO.sleep` fix](#) so terminal Vera programs run unchanged on both targets.

## Discovery context

An agent writing a browser variant of Conway's Life on current main reported this as the second of two related obstacles to "write once, run anywhere":

> *"ANSI escapes don't render because the browser runtime treats `IO.print` as text-to-DOM rather than text-to-terminal."*

Their working terminal program (`life.vera`) used `\u{1B}[2J` (clear screen), `\u{1B}[H` (cursor home), `\u{1B}[K` (erase to end of line) on each frame to overwrite the previous render in place. In the browser these escape sequences appeared as literal control characters in the page output, so each frame appended below the previous instead of overwriting — turning what should be an animation into a long stripey text dump.

The agent's workaround was a separate browser-shaped program (`life_web.vera`) that emits packed `0`/`1` strings to stdout, paired with a custom JavaScript canvas driver — bypassing both `IO.print` rendering and ANSI escapes entirely. That works but isn't portable and isn't ergonomic.

## Proposed scope

A minimal ANSI subset is sufficient for cursor-addressable terminal apps (the kind animations and TUIs use). The minimum useful set:

- **Cursor positioning**: `ESC[H` (home), `ESC[<row>;<col>H` (move), `ESC[<n>A/B/C/D` (relative move).
- **Screen / line clearing**: `ESC[2J` (clear screen), `ESC[K` (erase to EOL), `ESC[J` (erase below cursor).
- **Basic colors (optional first step)**: `ESC[3<n>m` / `ESC[4<n>m` (foreground / background 8-color), `ESC[0m` (reset).

The runtime maintains a virtual screen buffer (rows × cols of cells with text + style) and applies these escapes against that buffer. On each `IO.print`, after parsing escapes, the buffer is rendered as styled `<span>`s into a target `<pre id="vera-screen">` element.

`xterm.js` is the maximalist implementation (very large, full xterm fidelity). The proposed scope is much smaller — just enough for cursor-addressable Vera programs. ~200 lines of JS.

## Why this is high-leverage

- **Closes the rendering half of the terminal-vs-browser IO seam** without language changes. Combined with the JSPI-`IO.sleep` fix, a typical terminal Vera program (`IO.sleep` for pacing + ANSI for screen control) renders identically on both targets.
- **Worked example**: The Vera Conway's Life from PR #601's transcript would Just Work — same source, both targets, no JS driver.
- **Compatibility with existing programs**: terminal Vera programs that already use ANSI escapes (mostly examples in `examples/` like animations, simple TUIs) run as-is in the browser without modification.

## Acceptance

- The runtime parses and acts on at least the cursor-positioning + screen/line-clearing escape subset listed above.
- A program that does `IO.print("\u{1B}[2J\u{1B}[H")` followed by frame content renders correctly in the browser (frame N replaces frame N-1, no concatenated stripes).
- A test in `tests/test_browser.py` verifies a small terminal animation (e.g., a 5-cell counter that increments via cursor-home) renders with the expected final DOM state after N frames.
- The full Conway's Life program from PR #601 (`life.vera`, terminal version) runs unchanged on `vera compile --target browser` and animates correctly.

## Related

- [JSPI-driven `IO.sleep` issue](#) — the timing half of the same seam; the two together close the gap
- [#608](https://github.com/aallan/vera/issues/608) — the framing issue ("terminal-vs-browser IO seam")
- The agent's full design memo is posted as a comment on the umbrella issue: [#608 design memo](https://github.com/aallan/vera/issues/608#issuecomment-4399625656).  Key paragraph: *"Cursor positioning ([ESC] [H, [K, [2J), basic colours. Render to a `<pre>` element, not the whole page. This is bounded scope — maybe 200 lines of JS — and lets terminal-style programs render in the browser with no JS driver at all."*

aallan:
## Summary

Adds browser-target reliability to the stabilisation tier in `ROADMAP.md`. Two Conway's Life agent experiments on `--target browser` surfaced five concrete blockers, and the agent's design memo maps each obstacle to a runtime-only fix path. Promoting #609 (JSPI-driven `IO.sleep`) and #610 (ANSI subset interpreter) into the stabilisation tier commits to "browser-target Vera is something you'd actually use" before the agent-integration push.

## Why

The current stabilisation framing recognises two patterns from the v0.0.120–v0.0.138 bug-killing campaign: scale-dependent bugs (#596 stress harness) and walker-completeness gaps (#597 walker audit). Both are codegen-side concerns. A third pattern is now visible from the same evidence base — *browser-target reliability is approximate, not real* — and deserves the same treatment.

Specifics, all surfaced by the recent agent experiments:

- **Timing**: `IO.sleep` busy-waits and freezes the tab. Animation programs literally can't run. Fix: [#609](https://github.com/aallan/vera/issues/609) (JSPI / Asyncify).
- **Rendering**: ANSI escapes render as literal text. Cursor-addressable terminal programs can't render. Fix: [#610](https://github.com/aallan/vera/issues/610) (ANSI subset interpreter, ~200 lines of JS).
- **Marshalling**: JS can't pass `String` arguments to Vera functions. Forces all browser programs into a "compute everything upfront" pattern. Fix: [#603](https://github.com/aallan/vera/issues/603) (export `runtime.mjs` helpers).
- **Codegen gap A**: Five prelude combinators silently skipped on every WASM compile. Bug: [#604](https://github.com/aallan/vera/issues/604).
- **Codegen gap B**: String-returning function call in interpolation produces invalid WASM. Bug: [#602](https://github.com/aallan/vera/issues/602).

The first two are the high-leverage ones for the stabilisation tier — they're the difference between "browser-target Vera is theoretical" and "browser-target Vera is something you'd actually use." The other three are tracked in `KNOWN_ISSUES.md` already.

## Changes

- **Campaign-pattern list**: 2 → 3 patterns. Third pattern documents the browser-target seam with links to the umbrella issue ([#608](https://github.com/aallan/vera/issues/608)) and the five concrete blockers.
- **Stabilisation tier**: 3 → 5 items. Added #609 (JSPI sleep) as item 3, #610 (ANSI subset interpreter) as item 4, with mechanism notes. Demoted #595 (wasmtime-py malloc abort) from item 3 to item 5 — it's contingent on an upstream release we don't control, so it's the longest-tail item.
- **Agent-integration tier**: renumbered 4/5/6 → 6/7/8 to match.
- **"What moves when" gate**: "#4 starts when #1–#3 are closed" → "#6 starts when #1–#5 are closed".

## Test plan

- [x] `python scripts/check_doc_counts.py` — clean
- [x] No content changes outside ROADMAP.md (single-file edit)
- [x] All issue references (#609, #610, etc.) resolve to the correct existing issues

## Summary by CodeRabbit

* **Documentation**
  * Updated the project roadmap to document runtime stability improvements, including scale-dependent bug tracking, diagnostic enhancements, and browser compatibility work. The implementation roadmap has been reorganised with updated priorities for ongoing stabilisation efforts.


aallan:
Folding into [#601](https://github.com/aallan/vera/pull/601) — both PRs originate from the same agent-experiment session and touch independent files (no conflict).  The merge-coordination cost of two PRs consistently exceeds the review-clarity benefit.  ROADMAP commit cherry-picked onto #601's branch as `757435e`.

Issue #648, opened 2026-05-11
Cyclic type aliases crash codegen with RecursionError instead of clean E1xx diagnostic


aallan:
**Origin:** Discovered while implementing #633.

## The bug

A pair of mutually-recursive type aliases:

```vera
type A = B;
type B = A;

public fn id(@A -> @A)
  requires(true)
  ensures(true)
  effects(pure)
{
  @A.0
}
```

Passes `vera check` cleanly, then crashes `vera compile` with:

```
RecursionError at vera/codegen/core.py:628 in _type_expr_to_wasm_type
```

## Why

`_register_alias` in `vera/checker/registration.py` resolves aliases one at a time.  When `type A = B` is processed, `B` is not yet registered, so it resolves to a placeholder `AdtType("B", ())`.  Then `type B = A` finds `A` and chains to it.  No cycle is observable in the resolved types because the unresolved-import fallback masks it.

But codegen stores the raw AST `type_expr` (`vera/codegen/core.py:388` and `registration.py:26`), and `_type_expr_to_wasm_type` chases the alias chain through the AST, where the cycle is preserved.  Hence the recursion explosion.

## Recommended fix

A post-registration pass in the type checker:

1. After `_register_all` finishes in `vera/checker/registration.py`, walk every alias's AST `type_expr` chain.
2. Detect cycles using a `seen` set per alias.
3. On detection, emit `[E132] Cyclic type alias` with location, rationale, and fix suggestion.

This makes the codegen defensive cycle guards (added in #633) belt-and-braces.

## Acceptance criteria

- Cyclic alias program produces `[E132]` from `vera check` (not from `vera compile`).
- Diagnostic includes location of the originating declaration.
- A regression test exists in `tests/test_checker_*.py` (or wherever alias tests live).

## Pairs with

- #633 (codegen-side cycle guards — defensive)
- #630 (canonicalisation centralisation)

aallan:
Discovered while validating #633's reachability claim.  Cross-linked from #633.  The cycle guard added in PR #649 for #633 becomes belt-and-braces once the proper checker-side rejection here lands.

Issue #630, opened 2026-05-08
Cross-cutting: centralise canonical Vera-type-name resolution to close the #602 bug class structurally


aallan:
**Origin:** Eight distinct triggers of the same `i32_pair`-into-`i64` mismatch shape, accumulated across [#602](https://github.com/aallan/vera/issues/602), [#614](https://github.com/aallan/vera/issues/614), PR [#627](https://github.com/aallan/vera/pull/627), and PR [#629](https://github.com/aallan/vera/pull/629). Each one was fixed locally; each fix added one more isinstance-handler or one more inference site. The next trigger is a matter of when, not if. Time to fix the *shape*.

## The pattern

Every trigger has the same four-step shape:

1. Some inference site in `vera/wasm/inference.py` walks a `TypeExpr` to extract a canonical Vera type name string (`"String"`, `"Array"`, etc.).
2. The walk is *ad-hoc*: each site re-implements its own combination of (a) `RefinementType` unwrap, (b) alias-chain follow via `_resolve_base_type_name`, (c) generic substitution, (d) format-with-args.
3. Some combination is missed at some site → the walk returns a non-canonical string (`"Str"`, an alias name) or `None`.
4. Downstream `_translate_interpolated_string` (`vera/wasm/operators.py:455-486`) checks `vera_type == "String"` → mismatch → silent fallthrough to `to_string(...)` over `i32_pair` → `expected i64, found i32` at WASM validation, OR a `call_indirect` indirect-call type mismatch at runtime.

The cartesian product is *(N inference sites) × (M type-expression shapes)*. We've patched roughly 7 sites and 5 shapes — both numbers will keep growing as the language grows.

## Triggers fixed so far

| # | PR | Trigger shape | Site fixed |
|:---:|:---:|---|---|
| 1 | [#602](https://github.com/aallan/vera/issues/602) | Plain `String`-returning FnCall in interpolation | `_infer_fncall_vera_type` i32_pair branch (registry consult added) |
| 2 | [#614](https://github.com/aallan/vera/issues/614) | `f()[i]` where `f` returns `Array<T>` | `_infer_index_element_type_expr` FnCall branch |
| 3 | PR [#627](https://github.com/aallan/vera/pull/627) | Type alias over String (`type Str = String`) | Same site, `_resolve_base_type_name` resolution added |
[...]

Issue #589, opened 2026-05-06
host_print / host_stderr / host_contract_fail crash with raw Python UnicodeDecodeError on invalid UTF-8 bytes


aallan:
## Summary

A Vera program that calls `IO.print` with a `String` whose underlying bytes are not valid UTF-8 (typically because an upstream codegen bug corrupted a String's `(ptr, len)` pair) causes the runtime's `host_print` host import to raise an unhandled `UnicodeDecodeError`. wasmtime-py wraps this as a "python exception" cause and the full Python traceback escapes to the user's terminal — even though the program is valid Vera and the caller did nothing wrong.

A user-level program should NEVER produce a Python traceback. This violates the `WasmTrapError` contract established by #516 / #522 / #547 (every runtime trap is classified, carries a source backtrace, and surfaces a Vera-native fix paragraph).

## Severity

Critical UX bug. The user sees a 30+ line Python traceback ending in `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 123: invalid start byte` instead of a Vera-native runtime trap. The traceback exposes wasmtime-py internals (`enter_wasm`, `trampoline`, `maybe_raise_last_exn`) which are noise for the Vera user.

The trigger here is #588 (captured-Array indexing in closure produces silent string corruption that leaks into `IO.print`), but the right fix is independent: the host-print path must always handle invalid UTF-8 gracefully regardless of *why* the bytes are corrupt.

## Reproducer

Any program that triggers a codegen bug producing corrupt String pointers can hit this. The simplest known reproducer is the full Conway's Game of Life from issue #588:

```bash
$ vera run life_full_program.vera
[... ANSI animation runs for one generation ...]
Generation 0 of 200
Traceback (most recent call last):
  File "/.../wasmtime/_func.py", line 234, in enter_wasm
    yield byref(trap)
  File "/.../wasmtime/_func.py", line 103, in __call__
    raise WasmtimeError._from_ptr(error)
wasmtime._error.WasmtimeError: error while executing at wasm backtrace:
    0:    0xf0f - <unknown>!run_loop
    1:   0x1335 - <unknown>!run_loop
Caused by:
    python exception

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/.../vera/cli.py", line 1237, in main
    sys.exit(cmd_run(...))
  File "/.../vera/cli.py", line 600, in cmd_run
    exec_result = execute(...)
[...]

Issue #584, opened 2026-05-06
User-defined Unit-returning function call in non-tail position breaks WASM (block statement-sequencing emits drop with empty stack)


aallan:
## Summary

In a block expression, the pattern `helper(...); next_expr` where `helper` is a **user-defined function returning `@Unit`** in **non-tail position** produces invalid WASM.  Type-checks cleanly, then `vera run` fails at WASM instantiation with a stack-balance error.

Surfaces in any IO-effecting program of non-trivial size, because the natural `render(grid); IO.sleep(120); recurse(...)` shape lands on it the moment `render` is a user helper rather than a single inline `IO.print`.  Discovered by a sandboxed Claude.ai instance writing Conway's Game of Life — the same session as #583.

## Reproducer

```vera
effect IO {
  op print(String -> Unit);
}

private fn helper(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
  IO.print("middle ")
}

public fn main(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
  IO.print("start ");
  helper(());
  IO.print("end\n")
}
```

`vera check` → OK
`vera run`:
```
Error: Unhandled WASM trap: failed to compile: wasm[0]::function[2]::main

Caused by:
    0: WebAssembly translation error
    1: Invalid input WebAssembly code at offset 86: type mismatch:
       expected a type but nothing on stack
```

## Narrowing matrix

| Variant | Result |
|---|---|
| `IO.print("a"); helper(()); IO.print("c")` where `helper -> Unit` (user) | **Fails** |
| `IO.print("a"); IO.print("b"); IO.print("c")` (three built-ins) | OK |
| `IO.print("a"); helper(()); IO.print("c")` where `helper -> Int` (user) | OK |
| `IO.print("a"); helper(())` (Unit user fn in **tail** position) | OK |
| `helper(()); IO.print("c")` (no leading IO.print, Unit user still mid-block) | **Fails** |

So the trigger is specifically: user-defined function returning `@Unit`, called in a non-tail statement position within a block.  Built-in `IO.*` operations returning `Unit` work.  Returning anything else with a value rep works.  Putting the same Unit fn in the tail position works.

## Root cause

`vera/wasm/context.py::_is_void_expr` (line 456) determines whether an expression produces a WASM stack value.  For `ExprStmt`s in non-tail position, the surrounding `translate_block` (line 422-434) emits a `drop` instruction iff `_is_void_expr` returns False.

[...]

PR #649, opened 2026-05-11
Tier A bug burn-down: #633 + #634 + #556 + #591

### LINKED ISSUE #556: Calling a user-defined Unit-returning function as a statement fails WASM compilation
## Summary

Calling any user-defined function whose return type is `@Unit` in **statement position** (followed by `;`, with a separate final expression) traps `vera run` with `WebAssembly translation error: type mismatch: expected a type but nothing on stack`. `vera check` and `vera verify` both pass cleanly — the error only surfaces at WASM compile time.

The same call works when written as the block's final expression. Built-in `IO.print` (which also returns `@Unit`) is **not** affected, even in statement position.

## Minimal repro (13 lines, no IO, no ADTs, no String)

```vera
private fn pure_helper(@Nat -> @Unit)
{
  ()
}

  requires(true) ensures(true) effects(<IO>)
{
  pure_helper(1);
  ()
}
```

```
$ vera check repro.vera
OK: repro.vera

$ vera run repro.vera --fn main
Error: Unhandled WASM trap: failed to compile: wasm[0]::function[1]::main

Caused by:
    0: WebAssembly translation error
    1: Invalid input WebAssembly code at offset 59:
       type mismatch: expected a type but nothing on stack
```

## What works (control)

Replace `pure_helper(1);` with the call as the **final expression** and it compiles & runs:

```vera
  requires(true) ensures(true) effects(<IO>)
{
  pure_helper(1)        -- no `;`, no trailing `()`
}
```

A built-in `Unit`-returning call as a statement also works:

```vera
  requires(true) ensures(true) effects(<IO>)
{
  IO.print("hi");       -- statement form is fine for builtins
  ()
}
```

## Hypothesis (offered for triage; please disregard if wrong)

The codegen path for **user-defined function calls** in statement position seems to leave the `Unit` return value on the WASM stack (no `drop`), while the path for built-in `IO.*` operations either returns nothing or drops correctly. The next expression `()` then pushes a value the verifier didn't expect to find a slot for, surfacing as "expected a type but nothing on stack" once the function epilogue runs. A targeted check would be: emit `drop` after every statement-position call whose return type is `Unit` (or non-`Never`), regardless of whether the callee is built-in or user-defined.

## Why it matters

The natural pattern for breaking a `main` function into helpers is:

```vera
helper_a();
helper_b();
helper_c();
()
```

Right now this requires every helper to be inlined into `main` or chained as `(((helper_a(); helper_b()); helper_c()))`-style nesting, which defeats the point. SKILL.md's own "Best Practices" section recommends extracting helpers to control De Bruijn slot complexity, so this bug bites exactly the workflow Vera asks agents to follow. (Encountered while writing a 250-line λ-calculus interpreter; tried to factor a `report(@String, @Term, @Nat -> @Unit)` helper out of `main`; everything check'd and verify'd cleanly but `main` then refused to compile to WASM.)

## Environment

- Vera v0.0.127 (editable install from `aallan/vera@main`)
- Python 3.14.3, macOS 25.4 (arm64)
- `wasmtime-44.0.0`

### LINKED ISSUE #591: HTTP / Inference network-response decode uses strict UTF-8 (Python error noise leaks into Result Err)
## Summary

Three network-response UTF-8 decode sites in `vera/codegen/api.py` use strict-mode `bytes.decode("utf-8")`. If a remote API returns a response containing non-UTF-8 bytes, the user sees a Python `UnicodeDecodeError` message **leaked into a Vera-level `Result::Err` string** rather than something Vera-native and actionable.

This is a *lower-severity sibling* of #589 — the sites are all wrapped in `try/except Exception` → `_alloc_result_err_string`, so the failure does NOT escape as a raw Python traceback (the way #589 did via `host_print`). But the error message handed to the user contains Python-internals noise (`UnicodeDecodeError: 'utf-8' codec can't decode byte 0x... in position N`) which doesn't help them understand what went wrong with their `Inference.complete` / `Http.get` / `Http.post` call.

Found during the #590 (`/pr-review-toolkit:review-pr`) follow-up: I grepped for ALL strict-decode sites across the codebase to make sure #589's fix was complete. These three are the remaining ones; they're a different layer (network data, not WASM memory) so they were deliberately not bundled into #590.

## Affected sites

| File:line | Used by | Trigger |
|---|---|---|
| `vera/codegen/api.py:756` | `Inference.complete` (host_inference_complete) | Remote LLM API returns non-UTF-8 response |
| `vera/codegen/api.py:2809` | `Http.get` (host_http_get) | URL response body is non-UTF-8 |
| `vera/codegen/api.py:2841` | `Http.post` (host_http_post) | URL response body is non-UTF-8 |

## Why it's lower severity than #589

- The bytes come from a *remote* server, not from a Vera-program-internal codegen bug. A misbehaving remote API is a real failure mode that the user *should* know about.
- The `try/except Exception` wrapping means no Python traceback escapes — the user sees a `Result::Err(...)` they can match on, not a wasmtime-trampoline-wrapped Python crash.
- Practical trigger probability is very low: HTTP/JSON APIs almost universally use UTF-8 with explicit Content-Type. A misconfigured server is the only realistic source.

## Why it should still be fixed

- The error message contains Python-internals noise that the Vera user can't map back to anything actionable in their program.
- Setting `errors="replace"` (matching the #589 sites) would let the user see the actual response body with U+FFFD substitutions, surfacing API misbehaviour in a way they can debug.
- Alternative: detect invalid UTF-8 explicitly and surface a Vera-native `Err("response body was not valid UTF-8")` — preserves the failure signal but strips Python noise.

## Trade-off

The structured-error path (option B) preserves the diagnostic signal but loses the actual response bytes. The `errors="replace"` path (option A) loses the diagnostic signal (the user might not notice U+FFFD chars in their string) but preserves data. Most plausible answer: option A for `Http.get` / `Http.post` (user wants the data); option B for `Inference.complete` (where a non-UTF-8 response from an LLM API is genuinely broken and should fail loudly).

Worth deciding before fixing.

## Reproducer (synthetic)

Hard to reproduce naturally — would need a mock HTTP server returning `Content-Type: application/json; charset=utf-8` with bytes that aren't valid UTF-8. The structural concern is verifiable by inspection of the four lines without needing a runtime trigger.

## Acceptance

- All three sites use either `errors="replace"` or an explicit invalid-UTF-8 detection that surfaces a Vera-native `Err`.
- Existing structural tests in `tests/test_runtime_traps.py::TestHostPrintInvalidUtf8589` extended to cover these three sites for regression-pinning.
- No new bug introduced for valid-UTF-8 responses (which is the overwhelming common case).

## Related

- #589 — same defensive-hygiene class, fixed at the WASM-memory-decode layer.
- #588 — upstream codegen bug producing corrupt String pointers (the practical trigger of #589, irrelevant here).
- The PR #590 review comment (https://github.com/aallan/vera/pull/590#issuecomment-4390690759) deferred this site for separate analysis.

### LINKED ISSUE #633: _resolve_base_type_name recurses without cycle guard (post-#630 inconsistency)
**Origin:** [test-analyzer agent finding C1 / silent-failure-hunter M1 / code-reviewer S4 on PR #631](https://github.com/aallan/vera/pull/631).

## The pattern

`_resolve_base_type_name` in `vera/wasm/inference.py:~1340` is a recursive alias-chain follower with no cycle guard:

```python
def _resolve_base_type_name(self, name: str) -> str:
    if name not in self._type_aliases:
        return name
    alias = self._type_aliases[name]
    if isinstance(alias, ast.RefinementType):
        if isinstance(alias.base_type, ast.NamedType):
            return self._resolve_base_type_name(alias.base_type.name)
    if isinstance(alias, ast.NamedType):
        return self._resolve_base_type_name(alias.name)
    return name
```

The deleted helper `_resolve_type_name_to_wasm_canonical` (removed in #630) had explicit cycle protection via a `seen` set.  Its replacement `_canonical_named_type` (the new central walker) also has cycle protection.  But `_resolve_base_type_name` continues to recurse without one, and is called from ~9 sites across `inference.py` plus `vera/wasm/data.py` and `vera/wasm/context.py`.

## Reachability

Empirically verified during PR #631 review: a cyclic type alias (`type A = B; type B = A;`) crashes the codegen with `RecursionError` at `vera/codegen/core.py:_type_expr_to_wasm_type` *before* reaching `_resolve_base_type_name`.  So the immediate vulnerability is masked by a different recursive resolver crashing first.

This means: `_resolve_base_type_name` is *currently* dead-code-safe with respect to cycles.  But:

- If the type checker ever lets a cycle through and the upstream `_type_expr_to_wasm_type` is fixed first, this becomes the next crash site.
- The inconsistency itself is a code-smell — the new walker has cycle protection, the old helper doesn't, and they sit side by side.

## Recommended fix

Either:

1. **Add a cycle guard to `_resolve_base_type_name`** for defence-in-depth (5-line change, low risk):
   ```python
   def _resolve_base_type_name(self, name: str, seen: set[str] | None = None) -> str:
       if seen is None:
           seen = set()
       if name in seen or name not in self._type_aliases:
           return name
       seen.add(name)
       # ... existing logic, threading `seen` through recursive calls
   ```

2. **Migrate remaining callers onto `_canonical_named_type`** — strictly aligned with the #630 goal but a bigger surgery (the call sites use string→string, the new walker uses TypeExpr→NamedType).  Probably too large for a single follow-up.

Option 1 is the cheap defensive move.  File only if there's appetite.

## Verification needed

Before either fix, confirm whether the type checker rejects cyclic aliases.  If yes, this whole concern is purely belt-and-braces.  If no, the latent bug is real and Option 1 should land.

## Pairs with

- #630 (canonicalisation centralisation — landed; this is the residual cleanup).
- #626 (broader silent-skip pattern).

### LINKED ISSUE #634: SlotRef nodes inside InterpolatedString carry unreliable spans (E615 source-location quality)
**Origin:** Surfaced during PR #631's review pass while writing a source-location assertion for the new `[E615]` diagnostic test.

## The pattern

When `_translate_interpolated_string` records a failing segment in `WasmContext._interp_inference_failures` and the harvest in `CodeGenerator._compile_fn` emits an `[E615]` diagnostic, the diagnostic's source location reads from the failing AST node's `.span`.  For `SlotRef` nodes inside an interpolation segment (e.g. `"\(@Option<Int>.0)"`), the span is unreliable — the diagnostic line/column lands on adjacent syntax (e.g. the closing brace of an earlier construct, line 3 in the test fixture rather than the actual line 9 of the SlotRef).

This is a transformer / parser concern, not a codegen one — the AST node is being constructed with the wrong span by `vera/transform.py` or the upstream Lark tree.

## Why it matters

PR #630's Tier 2 framing is: "convert silent miscompilation into source-located diagnostic".  The diagnostic is loud — visible as `[E615]` — but its source location can be wrong, which undercuts the "source-located" half of the contract.  The `TestE615LoudInterpolationFallthrough630::test_e615_fires_on_adt_in_interpolation` test was softened during PR #631 review (assertion changed from "line ≥ 5" to "line > 0") to accommodate this; once fixed, the assertion can be tightened to actually verify the line points at the segment.

## Recommended fix

Audit how `SlotRef` nodes are constructed inside `InterpolatedString.parts`.  Likely candidates:

- The Lark grammar for interpolation segments may not propagate per-segment positions correctly.
- `vera/transform.py` may be using the parent InterpolatedString's span when constructing inner SlotRefs.

A reproducer test fixture exists in `tests/test_codegen.py::TestE615LoudInterpolationFallthrough630::test_e615_fires_on_adt_in_interpolation` (the softened assertion).  Tightening that assertion is the natural acceptance test.

## Out of scope for #630

#630's scope was the canonicalisation walker + the loud-skip mechanism, not span fidelity.  Span issues affect every diagnostic that points at AST nodes constructed in this way, not just `[E615]`.

## Pairs with

- #630 (loud diagnostic; this issue improves the diagnostic's quality).

aallan:
## Summary

Tier A bug burn-down — four bug closures in one PR. All small, all independent, all pure-quality fixes (no breaking changes, no new user-facing behaviour beyond `[E615]` diagnostic accuracy).

### Closes #633 — defensive cycle guard

`_resolve_base_type_name` (in `vera/wasm/inference.py`) now carries an explicit `_seen` cycle-detection accumulator, restoring consistency with the post-#630 `_canonical_named_type` walker that already had one. Defence-in-depth: cyclic type aliases are user errors that should be rejected upfront by the type checker (filed separately as **#648** — the upstream `RecursionError` observed while validating #633's reachability), but a bug in the upstream rejection must not turn into a stack-blow inside codegen.

### Closes #634 — SlotRef span fidelity

`SlotRef` and other AST nodes constructed inside `InterpolatedString.parts` now carry source spans in **original-source coordinates** instead of synthetic-wrapper coordinates. `_parse_interp_expr` previously wrapped each segment in a dummy `fn interpExpr() { <SEG> }` for parsing, placing the segment at wrapper line 3, col 3 — and Lark-emitted spans inside the parsed body were never translated back. The result: `[E615]` diagnostics on interpolated expressions landed on line 3 of the user's file regardless of where the string literal actually was.

Fix: `_split_interpolation` now records each segment's offset within the raw string; `string_lit` computes the segment's original line/column from the outer string-literal span; `_parse_interp_expr` walks the parsed AST and remaps every `Span` via a new `_remap_spans_inplace` helper.

Test assertions tightened: `test_e615_fires_on_adt_in_interpolation` now pins both line (8) **and** column (15) of the diagnostic at the SlotRef's position; `test_multiple_e615_in_one_interpolation` pins per-segment column fidelity (two SlotRefs in the same string get distinct columns 15 and 37 matching their actual source positions).

### Closes #556 — Unit-fn-in-statement-position regression pin

The user-visible bug class (calling a user-defined `@Unit`-returning function in statement position trips WASM validation with `type mismatch: expected a type but nothing on stack`) was already fixed by **#584** in v0.0.135 (`_is_void_expr` in `vera/wasm/context.py` now recognises user-defined `@Unit` fns via `_fn_ret_types`). The specific repro shape from the original #556 report — a *pure* helper (no IO effect) followed by a unit-literal final expression — wasn't pinned by the existing conformance test `ch07_unit_fn_nontail.vera` (which covered IO-effect variants only). Added `TestUserUnitFnInStatementPosition556` with two cases: the exact #556 repro plus the where-block variant from the follow-up comment.

### Closes #591 — HTTP / Inference UTF-8 hygiene

Three network-response decode sites in `vera/codegen/api.py` no longer leak Python `UnicodeDecodeError` text into Vera-level `Result::Err` strings. Two strategies, chosen per-site by user intent:

- **`Http.get` / `Http.post`** — `errors="replace"`. A remote server returning non-UTF-8 bytes surfaces as U+FFFD substitutions in the Ok-branch string, preserving the data. User intent for these calls is "fetch this URL"; preserving the body beats preserving the (already-rare) signal that bytes weren't cleanly UTF-8.
- **`Inference.complete`** — `_call_inference_provider` now catches `UnicodeDecodeError` explicitly and re-raises as `RuntimeError("Inference provider '<name>' returned a response body that is not valid UTF-8 (invalid byte at position N).")`. Non-UTF-8 from an LLM API is genuinely broken; we want loud failure with a Vera-native message, not the `codec can't decode byte 0x...` Python form.

Three structural assertions in `tests/test_runtime_traps.py::TestNetworkResponseUtf8Hygiene591`, mirroring the #589 coverage shape.

## What I discovered along the way

While verifying #633's reachability claim, I confirmed that cyclic type aliases (`type A = B; type B = A;`) currently pass `vera check` cleanly then crash `vera compile` with a `RecursionError` from `vera/codegen/core.py:_type_expr_to_wasm_type`. Filed as **#648** — the proper fix is a post-registration cycle-detection pass in `vera/checker/registration.py` emitting `[E132]` with location and rationale. The #633 cycle guard is belt-and-braces for that family.

## Burn-down context

Today's bug-burn-down: started at 99 open issues, this PR closes 4, files 1 (#648), and the upcoming spec docs PR will close ~3 more. Net trajectory: 99 → ~93.

## Test plan

- [x] `pytest tests/` — 3,802 passed, 14 skipped
- [x] `mypy vera/` — Success: no issues found in 58 source files
- [x] `python scripts/check_conformance.py` — All 86 conformance programs pass
- [x] `python scripts/check_examples.py` — All 34 examples pass (check + verify)
- [x] `python scripts/check_doc_counts.py` — Documentation counts consistent (3802 tests)
- [x] `ruff check --select S vera/` — All checks passed

Closes #633.
Closes #634.
Closes #556.
Closes #591.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented infinite recursion for cyclic type aliases during compilation.
  * Improved interpolation diagnostics so reported spans point to exact source locations.
  * Harmonised UTF‑8 handling so invalid bytes don’t leak internal codec errors into user-facing messages.

* **Tests**
  * Added regression tests covering cyclic aliases, interpolation span fidelity, and UTF‑8 decode hygiene.

* **Documentation**
  * Updated changelog, known-issues, testing metrics and roadmap counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aallan/vera/pull/649)


aallan:
@coderabbitai Both valid — fixed in b623aaf.

**Finding 1** (`_resolve_base_type_name` cycle guard): you're right that the prefix-chain shape `{A: B, B: C, C: B}` returned the revisited cycle node `B` instead of the caller's input `A`.  My initial test only covered full cycles (where caller-name == revisited-name by definition), hiding the bug.  Threaded a `_root_name` parameter through the recursion so the guard returns the caller's original input.  Added a prefix-chain test case as the regression pin.

**Finding 2** (`_split_interpolation` escape skipping): valid and a pre-existing bug — `_has_interpolation` correctly skipped escaped pairs but `_split_interpolation` didn't, so a string like `"a\\(b)c"` parse-crashed instead of resolving to a literal `a\\(b)c`.  Aligned the escape-skipping shape with `_has_interpolation` and added `test_escaped_backslash_before_paren_is_literal` as the regression pin.  Fixed opportunistically since I'm in the file and the two-helpers-must-agree invariant is exactly what this code is supposed to encode.

aallan:
@coderabbitai Valid finding on `test_http_post_uses_errors_replace` — fixed in c7fbf63.  Verified the bare substring `errors="replace"` appeared in **both** the explanatory comment block above the call site **and** the actual decode call, so a future regression removing the code while leaving the comment intact would have passed the test spuriously.

For `test_inference_complete_catches_unicode_decode_error`: the current code is **not** vulnerable (production `except` keyword does not appear in nearby comments — they use backtick-quoted `\`UnicodeDecodeError\``), but the substring assertion was structurally fragile against a future comment containing the literal `except UnicodeDecodeError`.  Hardened defensively under the same banner.

Both tests now use `re.search` with patterns that bind the target phrase to the surrounding code expression (`resp.read().decode([^)]*errors="replace"` for http_post, and `except UnicodeDecodeError ... raise RuntimeError(` block + raise's string argument check for inference).  Validated the regex rejects the comment-only-stub failure mode with a direct counter-example test.

Issue #372, opened 2026-03-27
Inference effect: user-defined handlers (handle[Inference])


aallan:
The `<Inference>` effect is currently host-backed only — there is no way to write a user-defined `handle[Inference]` block in Vera code. This prevents local mocking, replay testing, and routing to local models without going through the host.

## Use cases

- **Testing**: mock `Inference.complete` in unit tests without a real API key
- **Local models**: route to llama.cpp, Ollama, or any OpenAI-compatible endpoint via a custom handler
- **Replay**: record/replay LLM responses for deterministic tests

## Design note

User-defined effect handlers require the handler dispatch mechanism (currently only wired for user-declared effects). The built-in Inference effect would need to become handleable like `Http` and `IO`. This is the same limitation as the Http effect.

## Related

Deferred from [#61](https://github.com/aallan/vera/issues/61) (v0.0.101).

aallan:
### Tracked in compiler / spec limitation tables

This issue is referenced as an open limitation in: [`vera/README.md` § Current Limitations](https://github.com/aallan/vera/blob/main/vera/README.md#current-limitations).

When closing this issue, please update those references too — the limitation tables in `vera/README.md` and the spec chapters don't get the same regular hygiene pass as [`KNOWN_ISSUES.md`](https://github.com/aallan/vera/blob/main/KNOWN_ISSUES.md), so it's easy for them to drift behind reality.

aallan:
## Spec-update reminder for when this is fixed

When user-defined `handle[Inference]` handlers land, the PR that closes this issue must update the spec:

- **`spec/09-standard-library.md`** §Inference (currently around line 564) — the bullet "User-defined `handle[Inference]` handlers (for mocking, local models, replay) are planned for a future release ([#372])".  Remove the bullet and document the new shape (with examples) in the same section.

Added 2026-05-11 as part of [#650](https://github.com/aallan/vera/pull/650).

aallan:
Cross-link: #270 shares the root cause — a built-in host-backed effect (Inference here, Async there) cannot be intercepted by user handlers. Both need the same handler-dispatch mechanism for host effects.

aallan:
[...]

PR #700, opened 2026-05-28
docs: add inference handler mock example

### LINKED ISSUE #380: Add effect handler mocking example for Inference
Add an example showing `handle[Inference]` substituting a deterministic response for testing:

```vera
private fn with_mock_inference(@String -> @Result<String, String>)
  requires(true)
  ensures(true)
  effects(pure)
{
  handle[Inference] {
    complete(@String) -> { resume(Ok("Positive")) }
  } in {
    classify_sentiment(@String.0)
  }
}
```

This demonstrates the key architectural advantage of modelling inference as an algebraic effect: testability via handlers. Without this example, the mockability benefit is theoretical rather than demonstrated.

Relates to #372 (user-defined Inference handlers).

sarvesh1327:
## Summary
- Adds an `inference_mock.vera` example showing `handle[Inference]` with a deterministic `Inference.complete` response
- Documents the example in `examples/README.md` and lists the existing inference example in the Effects table
- Refreshes the examples count to match the current top-level example set

## Test Plan
- [x] `uv run --project . vera check examples/inference_mock.vera`
- [x] `uv run --project . vera verify examples/inference_mock.vera`
- [x] `uv run --project . python scripts/check_examples.py`

Closes #380

## Documentation
* Expanded available examples collection with two new inference-related demonstrations, bringing the total example count from 33 to 36.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aallan/vera/pull/700?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,6 @@
 # Vera Examples
 
-33 example programs demonstrating Vera's features. All examples pass `vera check` and `vera verify`.
+36 example programs demonstrating Vera's features. All examples pass `vera check` and `vera verify`.
 
 ## Running Examples
 
@@ -48,6 +48,8 @@ vera run examples/factorial.vera --fn factorial -- 10
 | `file_io.vera` | `vera run examples/file_io.vera` | File read/write with error handling |
 | `async_futures.vera` | `vera run examples/async_futures.vera` | Async effect, Future type, concurrent composition |
 | `http.vera` | `vera run examples/http.vera` | Http.get, JSON parsing, network I/O (requires network) |
+| `inference.vera` | `vera run examples/inference.vera` | Inference.complete, Result matching, API-key-backed LLM calls |
+| `inference_mock.vera` | `vera check examples/inference_mock.vera` | Inference effect handlers, deterministic mock responses, API-key-free check/verify workflows |
 
 ### Data Structures
 
--- a/examples/inference_mock.vera
+++ b/examples/inference_mock.vera
@@ -1,0 +1,28 @@
+-- Inference handler mock example: deterministic LLM responses for tests.
+-- Demonstrates: Inference.complete, algebraic effect handlers, resume.
+--
+-- `classify_sentiment` normally requires the Inference host effect. The
+-- wrapper below handles that effect locally and resumes the computation with a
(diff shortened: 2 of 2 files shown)


aallan:
Hi @sarvesh1327 — thank you for this contribution. You wrote a clean, well-structured example that does exactly what #380 asks for, and the test plan is thorough.

Unfortunately we need to close this PR for now — and the reason is on us, not you.

The example demonstrates `handle[Inference]` (a user-defined handler for the built-in Inference effect). The static stages of the compiler (parser + type-checker + verifier) accept the syntax, which is why your test plan (`vera check`, `vera verify`, `scripts/check_examples.py`) passes cleanly. But **codegen does not currently support `handle[Inference]`** — running `vera run` or `vera compile` on the file produces:

> `Function 'main' body contains unsupported HandleExpr: only State<T> and Exn<E> handlers supported (got Inference) — function skipped.`

This is tracked as #372 (user-defined `handle[Inference]` support), which is open and not yet implemented. So while your example checks and verifies, it doesn't compile or run end-to-end. We can't ship it as an example in `examples/` because every example there must pass `vera run` for users to learn from.

**This is our fault for not signposting clearly.** #380's body said only "Relates to #372", which was much weaker language than the actual blocker. A contributor reading #380 had no way to know the underlying feature wasn't shipped.

We've gone back and addressed this across the project today:

- Mapped the dependency graph across all 92 open issues
- Identified 7 hard blocking relationships (including #380 → #372)
- Used GitHub's native [issue relationships](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues) feature to mark each one — visible in the Relationships panel on each issue, queryable via GraphQL
- Created a `blocked` label and applied it to the affected issues, so they're filterable via `is:open label:blocked`
- Posted a follow-up comment on #380 noting the dependency on #372

So the exact trap that caught you here shouldn't catch the next contributor.

When #372 lands, your work is exactly the right shape to ship — we'll credit you in the example header and `CHANGELOG.md` entry when it does. If you're interested in compiler-level work, #372 itself is a meaty but well-scoped issue.

Thank you again for the careful contribution, and apologies for the wasted effort.

Issue #421, opened 2026-03-29
Decompose vera/codegen/api.py (2,228 lines): extract memory layout and execution runtime


aallan:
## Background

`vera/codegen/api.py` has grown to **2,228 lines** and now mixes three distinct concerns: memory layout utilities, the public compile/execute API, and the host runtime setup. It follows the same growth pattern as `codegen.py` before it was split in #155.

### `vera/codegen/memory.py` (~150 lines)
Extract the ADT memory layout utilities that are needed by both the compiler and the runtime:
- `ConstructorLayout` dataclass
- `_wasm_type_size()`
- `_wasm_type_align()`
- `_align_up()`

These are pure data/math utilities with no runtime dependency and are already imported by multiple codegen modules.

### `vera/codegen/runtime.py` (~800 lines)
Extract the wasmtime execution layer:
- `ExecuteResult` dataclass
- `_VeraExit` exception
- `execute()` — sets up the wasmtime `Store`, linker, memory, all host imports (IO, Http, Inference, Map, Set, Regex, Markdown, …)
- All `host_*` functions (currently ~1,400 lines of the file)

### Keep in `vera/codegen/api.py` (~300 lines)
- `CompileResult` dataclass
- `compile()` — the public compilation entry point

## Acceptance criteria

- All 3,157 tests pass
- `mypy vera/` clean
- Public API (`from vera.codegen.api import compile, execute, CompileResult, ExecuteResult`) unchanged
- `api.py` under 400 lines after the split

aallan:
Scope supersession (June 2026 sweep): the file has doubled since this was filed — `vera/codegen/api.py` is now **4,253 lines**, with `execute()` spanning lines 837–4253 (3,417 lines, 154 nested host closures). The June 2026 audit recommends extracting a `vera/runtime/` **package** (one module per host-binding family) rather than the single `runtime.py` this body sketches, and the planning rework (PR #740) sequences two prerequisites first: the characterization harness (#734) and the GC bucket-as-truth migration (#706) — decomposing before #706 would churn the exact code it rewrites. Roadmap Tier 1. The acceptance criteria here otherwise stand: public API unchanged, `api.py` small after the split.

Issue #387, opened 2026-03-27
Add mutation testing with `mutmut` for type checker and verifier


aallan:
Mutation testing answers: do the tests actually catch bugs, or do they just execute code? The 96% coverage metric tells you how much code the tests reach, not how well they catch errors.

Run `mutmut` against critical modules first:

```bash
pip install mutmut
mutmut run --paths-to-mutate=vera/checker/ --tests-dir=tests/
mutmut results
```

Start in detection-only mode (don't enforce kill rates). Focus on the type checker and verifier modules where a surviving mutant indicates a silent correctness bug.

aallan:
Broadening this (folding in #784, closed as a duplicate): the scope should be the **whole suite — 4,000+ unit tests AND the 92 conformance programs — with the conformance suite first**, not just `mutmut` on checker/verifier.

**Why now:** the #680 shadow/projection audit ([#778](https://github.com/aallan/vera/pull/778)) ran a 14-mutation sweep over a 57-test battery and found **8 tests green for the wrong reason** — they asserted `div_zero → tier3` but, lacking a stale outer, the untranslatable `let` produced no tracked shadow, so the divisor was `None` and took the simple `divisor is None → tier3` path, never engaging the shadow machinery. Breaking that machinery left them green. A normal green run can't tell a strong test from a weak one; only mutation testing can.

**Conformance first:** the 92 conformance programs are the spec-conformance backbone — a weak one means a language feature isn't actually validated. Mutation-test = mutate the compiler logic the feature relies on and confirm the program's declared level (`check`/`verify`/`compile`/`run`) now fails. A program that still passes its level under a relevant mutation isn't testing what it claims.

**Approach:** evaluate `mutmut` / `cosmic-ray` (auto-generated source mutants → *surviving-mutant* report) alongside the bespoke hand-targeted harness proven in #778 (back up source, replace a unique substring to break one component, run the suite, collect which tests flip RED, restore). Detection-only first; a periodic (not per-commit — it's slow) job or a mutation-score budget keeps it honest.

**Phases:** (1) harness + conformance suite; (2) unit suite module by module (the original `mutmut`-on-checker/verifier here is a concrete Phase 2 start); (3) optional CI/budget to catch new weak tests.

aallan:
### Worked example: a weak test, the mutation that exposes it, and the robust replacement

All three snippets below were run against the real compiler; the statuses are actual output, not illustrative.

[...]

Issue #680, opened 2026-05-18
Auto-inject obligations for primitive operations (division, modulo, array index) to match README's static-safety claim


aallan:
## Context

The README's lead example (line 38) states:

> Division by zero is not a runtime error — it is a type error. The compiler checks every call site to prove the divisor is non-zero.

The implementation does not match this unqualified claim. Given:

```vera
public fn divide(@Int, @Int -> @Int)
  requires(true)
  ensures(true)
  effects(pure)
{
  @Int.1 / @Int.0
}
```

`vera verify` reports `OK: divide.vera. Verification: 2 verified (Tier 1)` — both trivial contracts pass and the unguarded division stands. The trap fires at runtime, not at verify time.

The same applies to `%` (modulo) by zero, array indexing past the end, and any other primitive operation whose well-definedness depends on operand values.

The mental model in user-facing docs is "if it compiles, the divisor is provably non-zero." The actual mental model is "the verifier checks the contracts you wrote. If a primitive operation can trap and you have not guarded it with a contract or refinement type, the verifier will not synthesise a guard for you."

## What this issue tracks

The **implementation work** of making the docs claim true: auto-inject latent preconditions on primitive operations.

For every `@Int.a / @Int.b` in a function body, the verifier should add `@Int.b != 0` as an implicit obligation discharged from the function's precondition or by Tier 1 reasoning. Same for `%`, array indexing, and any other potentially-trapping primitive.

The docs-only fix (softening the README/spec language to match the current implementation) will land in a separate immediate docs PR. This issue tracks the ambitious version.

## Scope

Primitive operations whose well-definedness depends on operand values:

| Operation | Implicit obligation |
|---|---|
| `a / b` (Int/Nat) | `b != 0` |
| `a % b` (Int/Nat) | `b != 0` |
| `arr[i]` | `0 <= i && i < array_length(arr)` |
| `string_at(s, i)` | `0 <= i && i < string_length(s)` |
| `a - b` (Nat) | `a >= b` (already handled by #520) |

## Related

- #427 — Tier 2 verification (this work is Tier-2-adjacent; the obligation discharge often requires reasoning beyond the current Tier 1 decidable fragment, particularly for array indexing where the index expression involves a function call)
- #520 — `@Nat` subtraction underflow obligation (the existing precedent for this pattern; the codegen guard + verifier obligation discharge is already in place for Nat subtraction)
[...]

PR #786, opened 2026-06-22
test(734): characterization harness for execute() ahead of the #421 decomposition

### LINKED ISSUE #734: Characterization harness for execute() ahead of the runtime decomposition
## Problem

`execute()` (`vera/codegen/api.py:837-4253`) is now 3,417 lines containing 154 nested host-binding closures, in a file totalling 4,253 lines — roughly double the size recorded when #421 was filed. It is the single highest-risk decomposition target in the codebase, and it has **no characterization tests pinning its observable contract**: existing coverage exercises it only indirectly, through language-feature tests.

## Plan

Before any decomposition work (#421), pin the function's behaviour with a direct harness:

- One test per `ExecuteResult` field — `value` (each return shape: int / float / str / heap pointer / None), `stdout`, `state`, `exit_code`, `stderr` — crossed with the three completion modes: normal return, WASM trap, interrupt/exit.
- Pin the positional-constructor compatibility shape (`value, stdout, state, exit_code`) that the dataclass comment promises to external callers.
- Pin `capture_stderr=True` versus the default.

This harness is what makes the #421 / `vera/runtime/` package split reviewable: every extraction PR must keep it green.

## Sequencing

Per the 2026-06-10 audit (§0.4, §2.2): this harness **and** the GC bucket-as-truth migration (#706) land before decomposition begins — decomposing first would churn the exact code #706 needs to rewrite.

Size: M.

aallan:
## What

Adds `tests/test_execute_characterization.py` — a consolidated characterization harness pinning `execute()`'s observable contract, as the regression gate for the #421 `vera/runtime/` decomposition.

`execute()` (`vera/codegen/api.py`, a ~3,400-line body with 154 nested host-binding closures in a 4,358-line file) is the highest-risk decomposition target in the codebase. #421 will split it one host-binding family per PR; this harness is the single green gate every extraction PR must keep — a reviewer can read it top-to-bottom and know that if it stays green, the externally visible behaviour of `execute()` was preserved.

## The contract pinned

Every `ExecuteResult` field × the three (asymmetric) completion modes:

| Field | normal return | WASM trap | interrupt / exit |
|---|---|---|---|
| `value` | int / float / str / heap-ptr / None | — | None |
| `stdout` | captured, in order | preserved on the exception | preserved |
| `state` | final snapshot | — | preserved |
| `exit_code` | None | — | n (`IO.exit`) / 130 (Ctrl-C) |
| `stderr` | gated by `capture_stderr` | preserved | preserved |

The trap mode is the key asymmetry: it **raises `WasmTrapError`** (carrying a classified `kind` plus the output written before the trap) — it does not return a result. Plus two cross-cutting pins: the positional-constructor compatibility shape (`value, stdout, state, exit_code`) and `capture_stderr` True-vs-default.

## Reuse, not duplication

The issue assumed coverage was "only indirect," but ~12 direct `execute()`-level tests already existed — scattered across `test_codegen.py` + `test_runtime_traps.py`, embedded in feature classes, and several missing the discriminating assertion. The harness reuses those fixtures verbatim (each cross-referenced in an `# overlaps` comment) and adds what they lacked: `value is None` on `IO.exit`, the declared field order, the str-decode-vs-array-pointer asymmetry pinned side-by-side, `0 != None` on exit codes.

## Mutation-validated (no green-for-the-wrong-reason tests)

Per the weak-tests discipline (#387): each cell was confirmed to flip RED when its target return path in `api.py` is deliberately broken. 9 mutations applied, 9 caught, 0 leaks — the source file is restored after each:

| Mutation in `api.py` | Caught by |
|---|---|
| normal `value=value` → `None` | value-shape (int/float/str/array) + initial_state |
| normal `stdout` → `""` | stdout captured / order-preserved |
| normal gains `exit_code=0` | exit_code-is-None-on-normal-return |
| `IO.exit` `exit_code` → `None` | exit-code + zero-is-distinct-from-None |
| normal `state` → `{}` | state-reflects-final-put |
| trap `raise` → `return` | all three trap-mode tests (DID NOT RAISE) |
| Ctrl-C `exit_code=130` → `0` | keyboard-interrupt-maps-to-130 |
| normal `stderr` → `""` | stderr-captured-when-requested |
| dataclass field reorder | positional-shape + field-order |

## Scope

Test-only — no `vera/` or `spec/` change, no version bump, no CHANGELOG (the gate stays inert; `git diff --name-only` is `tests/` + `TESTING.md` + `README.md` + `ROADMAP.md`). This is the first Tier 1 item and **does not cut a release** — it folds into a combined release with #421 (next). #734's ROADMAP row is removed (done); its HISTORY entry rides the combined release.

Closes #734

## Summary by CodeRabbit

* **Documentation**
  * Updated project status and roadmap/testing progress metrics to reflect 4,727 total tests.
  * Expanded testing documentation with a new execution characterization entry covering value/stdout/state/exit-code/stderr behaviour, including trap and interrupt handling.

* **Tests**
  * Added a new pytest characterization suite that pins the externally observable `execute()` contract across normal completion, WASM traps, and interrupt scenarios, including strict result-field shaping and output capture rules.


aallan:
## Review round 1 — pr-review-toolkit + CodeRabbit

**pr-review-toolkit** (three agents — test-coverage, comment-accuracy, code-review) found **no critical or important issues**. The test-coverage agent independently re-ran a sample of the nine mutations and confirmed each flips RED, so the mutation-validation claim holds rather than being merely asserted. **CodeRabbit** generated **no actionable comments** (its Linked-Issues, Out-of-Scope, and Title checks all passed).

The low-severity polish they surfaced is addressed in `31613ad`:

| Finding | Raised by | Fix |
|---|---|---|
| Out-of-scope fields (`host_store_sizes` / `peak_heap_bytes`) cited to `test_stress.py`, but they're actually pinned in `test_codegen.py` | all three agents | corrected the docstring citation |
| `test_codegen.py:LINE` references will drift (one was off-by-one) | two agents | switched to `::test_array_wat_has_bounds_check` name citations (the form the rest of the file already uses) |
| "154 nested closures" drifts as `api.py` grows ahead of the split | one agent | softened to "150+" |
| Only three trap kinds exercised, without saying why | test-coverage agent | added a docstring note — the other three (`out_of_bounds` / `stack_exhausted` / `overflow`) aren't reliably producible through a real `execute()` end-to-end and are unit-pinned in `test_runtime_traps.py` against a synthetic `_FakeTrap` |
| Docstring coverage 4.35% | CodeRabbit pre-merge check | one-line docstrings on all 22 test methods, matching the existing test-file convention |

All of it is comment/docstring polish — the harness is behaviourally unchanged (22/22 green, mutation-validation intact). `TESTING.md` line count synced to 467.

aallan:
@coderabbitai full review

The 22 test-method docstrings landed in `31613ad`, but the Docstring Coverage pre-merge check still reports the pre-docstring 4.35% (1/23) — it looks like it didn't recompute on the incremental re-review. Requesting a full review so the pre-merge checks reflect the current head.

aallan:
@coderabbitai Good catch — the 15-vs-16 mismatch is real, but the fix direction is the reverse of the suggestion.

Verified against `pytest tests/test_conformance.py -rs`: the true skipped count is **16** (8 verify-only + 8 check-only run/verify skips). So line 9's "16" is correct, and the **skip table plus the line-130 prose were the stale side** — the table was missing `test_run[ch04_primitive_obligations]`, the verify-level conformance program added in #680 whose `run` stage auto-skips. Lowering line 9 to 15 / the total to 4,726 would have made the doc inconsistent with the live count and with the `check_doc_counts` CI gate, which passes at 4,727.

Fixed in `8c56131`: added the missing `ch04_primitive_obligations` row and bumped the line-130 prose from 15 to 16. The skip table now totals 16 rows, matching line 9 and the live `pytest` run.

Issue #747, opened 2026-06-16
Generalize @Nat narrowing obligation to projection binding sites (ADT sub-pattern, non-literal destructure)


aallan:
## Summary

#552 generalised the `@Nat >= 0` invariant obligation to binding sites where an `@Int` value narrows into a `@Nat` slot, covering `let`, call arguments, effect-operation arguments, *concrete* constructor fields, top-level match binds, and literal-tuple destructures (Tier-1 static), plus a runtime guard at the `let` site.

Six narrowing shapes remain uncovered:

1. **ADT sub-pattern binds** — `match opt { Some(@Nat.0) -> ... }` where `opt : Option<Int>` binds the `@Int` payload as `@Nat`. The bound value is a field projection whose source type the verifier does not retain (`slot_env` maps slot → Z3 var, not → Vera `Type`).
2. **Non-literal tuple destructure** — `let Tuple<@Nat, @Nat> = f()` where `f` returns a tuple of `@Int`. Same projected-component-type problem. (The verifier now invalidates the destructured slots with fresh vars so a *later* obligation stays sound — PR #748 — but the destructure narrowing itself is still not statically obligated.)
3. **Generic constructor fields instantiated to `@Nat`** — `Some(@Int.0)` building an `Option<Nat>` (or `Ok` / `Err` / a user generic ADT). The field type is a `TypeVar`, so `_is_nat_type(field_ty)` is False and the construction-site narrowing is skipped; recognising it needs the call-site instantiation (the expected type), which the verifier does not re-derive.
4. **Imported ADT constructors with concrete `@Nat` fields** — a public constructor from an imported module (`import foo(MkBox)` where `MkBox(@Nat)`) narrowing an `@Int` argument. `_register_modules` harvests only public *functions* into `self.env`, so `_lookup_constructor_info` cannot resolve the imported constructor's field types. Closing it needs harvesting imported public ADTs into the verifier's registries (which also triggers their SMT registration — a broader cross-module-verification change). Raised in the PR #748 review.
5. **Generic effect-operation formals instantiated to `@Nat`** — a generic effect `E<T>` with `op wait(T -> Unit)` used as `effects(<E<Nat>>)`. After PR #748, `_register_effect` populates each op's `OpInfo`, but the formal `T` is a `TypeVar`, so `_is_nat_type(formal)` is False and `E.wait(@Int.0)` is skipped — the same instantiation problem as item 3, here on the effect row. Raised in the PR #748 review.
[...]

Issue #746, opened 2026-06-16
Implement refinement-type predicate verification


aallan:
## Summary

User refinement-type predicates (e.g. `type PosInt = { @Int | @Int.0 > 0 }`) are not verified anywhere in the compiler. `RefinedType` (`vera/types.py:53`) preserves the predicate as an opaque AST node ("kept opaque — not verified in C3"), and the verifier strips `RefinedType` to its base at every site (`vera/smt.py:1102`), discarding the predicate. A value violating a refinement predicate is accepted silently — `tests/conformance/ch02_refinement_types.vera` exercises parsing/aliasing, not enforcement.

`@Nat` is a distinct primitive with a hardcoded `>= 0` invariant (`declare_nat`), not a `RefinedType`, so #552's binding-site obligation does not generalise to arbitrary predicates.

## Scope

Implement refinement-predicate verification end to end:

- Translate a `RefinedType` predicate to Z3, substituting the refinement's binder variable with the value being refined (a `substitute_predicate(pred, value) -> z3.BoolRef` helper).
- Discharge the predicate as a proof obligation at every site that introduces a refined value: function parameters, return positions, `let` bindings, constructor fields, and `@Int`-into-refinement narrowing (the refinement analogue of #552).
- Emit a runtime guard fallback where Z3 cannot discharge statically, and a diagnostic with a counterexample on violation.
- Conformance + unit tests asserting both discharge and violation.

## Context

Surfaced while scoping #552 (@Nat binding-site narrowing). #552 was deliberately scoped to the built-in `@Nat >= 0` invariant; general refinement-predicate verification is this separate, larger piece (the C4 milestone).

Issue #732, opened 2026-06-12
Per-monomorphization static verification for generic functions


aallan:
## Problem

`_verify_fn` skips all static verification for generic (`forall<T>`) functions — an early return records E520 warnings for non-trivial contracts and moves on. #555 documents the sharpest consequence (an unguarded `@Nat - @Nat` inside a generic body never produces E502 at verify time), but the gap covers every obligation class: contracts, the underflow walkers, and invariant checks all bypass generic bodies. The runtime guards still fire per monomorphization, so this is a silent *downgrade* from Tier 1 to Tier 3 rather than a soundness hole — but it is exactly the kind of quiet weakening the project's verification pitch cannot afford.

## Decision

Verify each **concrete instantiation** through the existing Tier-1 pipeline (decided 2026-06-12, weighed against implementing Tier 2 #427 first):

- Codegen already monomorphizes generics per concrete type instantiation (`vera/codegen/monomorphize.py`). Verification reuses the same substitution: for every instantiation reachable in the program, substitute concrete types into the generic function's AST and run the standard `_verify_fn` on the result.
- No type-variable SMT is needed — Z3 only ever sees concrete sorts, so the entire existing obligation machinery (including the #520/#551 walkers and the obligation cache, which dedupes repeated instantiations at the same types) applies unchanged.
- Diagnostics must name the instantiation, e.g. `in foo<Int> (instantiated at main.vera:12)` — a generic author needs to know *which* concrete typing failed.

Trade-offs accepted: verification cost scales with instantiation count (the obligation cache bounds duplication); **uninstantiated generic functions remain statically unverified** — they keep today's E520 warning, a documented limitation rather than a silent one.

## Relationship to Tier 2 (#427)

#427 stays on the milestone horizon as the symbolic-generic upgrade. When it lands, per-monomorphization results become its **differential oracle**: symbolic verification of `foo<T>` and per-instantiation verification of `foo<Int>`, `foo<String>`, … must agree, and any disagreement is a bug in one of the two.

## Acceptance

- The #555 acceptance program (generic `@Nat -> @Nat` with an unguarded `@Nat.0 - 1`) fails `vera verify` once instantiated at any concrete type.
- `vera verify --json` tier counts include monomorphized obligations.
- Uninstantiated generics keep the existing E520 warning behaviour.
- Conformance test for the verified-generic happy path.

Supersedes the static half of #555 when it lands; #551's `@Byte` sibling is covered by the same mechanism.

[...]

Issue #552, opened 2026-04-28
Generalize @Nat invariant check to all binding sites (broader generalisation of #520)


aallan:
The verifier currently checks the `@Nat >= 0` invariant only at function return positions (`test_int_to_nat_negative_caught`: `fn bad(...) -> @Nat { -1 }` correctly fails verification) and via implicit Z3 constraints on `@Nat` parameters (`test_nat_constraint_used`: `@Nat.0 >= 0` is asserted automatically). It does NOT check *intermediate* binding sites where an expression of broader type narrows into an `@Nat`-typed slot.

Concretely, after #520 + #551 close the unsigned-subtraction underflow holes, the remaining gap is:

```vera
private fn leaky(@Int -> @Int)
  requires(true)
  ensures(@Int.result == @Int.0)
  effects(pure)
{
  let @Nat = @Int.0 - 100;  // could be negative if @Int.0 < 100
  @Int.0                     // returns the parameter, not the bad let
}
```

The bad `let @Nat = ...` slips past the verifier because the `@Nat`-invariant check fires only at the function's *return* position. The let value's narrowing from `@Int` to `@Nat` is not obligation-checked. If the bad slot is later used in arithmetic, postcondition checks on the *result* might surface the violation — but only if the postcondition happens to depend on that path. Sites where the bad value silently flows into another expression propagate the invariant violation invisibly.

This is the architectural generalisation of #520 (subtraction is one specific site that violates the invariant) and #551 (`@Byte` subtraction).

## Scope

Generalize the existing return-position invariant check to fire at every binding site that introduces an `@Nat`-typed slot from a broader-typed value:

- Let bindings: `let @Nat = <Int expr>`
- Function call arguments: `f(<Int expr>)` where `f`'s formal parameter is declared `@Nat`
- Match arm pattern bindings narrowing to `@Nat`
- Other narrowing-to-`@Nat` sites the audit surfaces

Same Tier-3 fallback story as #520: when Z3 discharges the obligation, the binding compiles cleanly. When it cannot, codegen emits a runtime guard (`if (val < 0) trap underflow`) at the binding site.

## Why follow-up rather than bundled with #520

#520 is scoped specifically to *subtraction* — the most common and most demonstrably-broken path. Path B is a verifier-architecture generalisation — broader scope, broader test surface, broader audit of every place a fresh `@Nat` slot enters scope. Splitting keeps each PR reviewable and lets the subtraction fix ship while the architectural piece gets thought through.

## Open design questions

[...]

PR #788, opened 2026-06-23
Harden type-checker tests to assert diagnostic error codes (#387 mutation testing)

aallan:
## Phase 3 (fix-core): soundness-core test hardening

Following the #387 baseline sweep (80.8% caught, 2,038 survivors), this stage fixes the tests that **stayed green when the code was deliberately broken** — the weak tests a mutation score exists to surface. The goal is not to lift a number but to make the suite go **RED when behaviour breaks**.

Every killing test asserts a **specific observable** — the stable diagnostic `error_code`, the `severity`, or a type-propagation result — never free-text message prose (killable only by brittle exact-string assertions, the anti-pattern this avoids). Each cluster was confirmed by a targeted `mutmut` re-sweep (e.g. `_check_constructor_call` 63 → 26, `_check_handle` 74 → 51).

### 236 mutants killed across 10 commits

| module | clusters | killed |
|---|---|---|
| `obligations/` | `__getattr__` lazy exports, `verify_source` ok/file-propagation | 13 |
| `checker/control.py` | exhaustiveness E310–E313, match E302, handler E330–E335, patterns E320–E322 | 73 |
| `checker/calls.py` | constructor/call E210–E215, E220; op E203–E204, E240–E241; module E230–E233 | 92 |
| `checker/core.py` | data invariant E120 | 19 |
| `checker/expressions.py` | unary E146, index E160–E161, assume E173, old/new E174–E175 | 39 |
| **total** | | **236** |

The pattern throughout: existing tests exercised the right path but asserted only a prose substring (`"requires Bool operand"`), leaving the `error_code` literal free to mutate; strengthening them to assert the stable code (plus `severity` for warnings) kills the whole family. New tests fill genuine gaps (e.g. data-type invariants had zero tests).

### Why this stops at the checker: the verifier subsystem is at its practical mutation ceiling

I probed smt.py and verifier.py (the remaining ~1,284 survivors). The same technique doesn't land: the error-code surface is exhausted (E521/E523/E524 are degenerate or `# pragma: no cover`), a correct array-literal value differential killed **0** new survivors (the #777 index battery already covers that surface — I reverted the redundant tests), and every high-survivor verifier method already has a dedicated mutation-aware battery (#520, #552, #680, #732, #746, #747, #777). The residual survivors are `_report_*` diagnostic prose, defensive fallbacks the comments themselves mark unreachable, and redundant guards — genuine equivalents, not weak tests. Full write-up on #387; the deeper smt.py Z3-soundness audit is #392's mandate.

### Notes
- **Refs #387** — does **not** close it. #387 stays open for the whole-corpus sweep gated on #421.
- Not a release. CHANGELOG `[Unreleased]` carries a running bullet.
- Test-only; no behaviour change. Full suite green; `mypy` / `ruff` / doc-counts clean.

## Release Notes

* **Tests**
  * Hardened checker tests to validate stable, observable diagnostic outcomes (error/warning codes, severities, and payload details) across more scenarios.
  * Improved verification-session coverage for strict `ok` semantics, diagnostic source-file propagation, and safe verification of uninstantiated generics.
  * Added/expanded tests for the lazy-export shim, including correct resolution of exported session objects and expected error handling for unknown attributes.

* **Documentation**
  * Updated testing guidance and project metrics, and refreshed changelog/roadmap entries for the revised suite totals.


aallan:
## Review round — pr-review-toolkit + CodeRabbit

**pr-review-toolkit** (pr-test-analyzer + code-reviewer) surfaced one real issue and two minor ones; all fixed in `715cdfd`:

1. **CI order-fragility (blocker, fixed).** Ten strengthened tests asserted `errs[0].error_code == "EXXX"` positionally. `_check_err` returns *all* error diagnostics unsorted, and cross-module global state (the ADT-registry resync `TESTING.md` itself flags) can leak a diagnostic ahead of the target — the code-reviewer reproduced a `1 failed` full-suite run for `test_bool_missing_carries_e312` (GitHub CI happened to pass on a lucky ordering). Converted all ten to the order-independent `any(e.error_code == "EXXX" for e in errs)` pattern this PR already uses elsewhere; that also picks up the co-emitted `E335` in the `E334` slot-mismatch test.
2. **False kill-claim corrected.** `test_uninstantiated_generic_verifies_without_crash`'s comment claimed to kill the `_instances.get(name, set())` default, but that default is unreachable (`_collect_instantiations` pre-seeds a `set()` for every generic) — an equivalent mutant, not a coverage gap. Reworded to an honest smoke-test note.
3. **CHANGELOG accuracy.** Added the omitted body-type `E121` to the `[Unreleased]` code list.

**CodeRabbit:** clean across both heads — 0 actionable inline comments, 0 out-of-diff findings, 4 checks passed. Its one warning is *Docstring Coverage 77%*, which is the **file-level legacy** figure: `tests/test_checker.py` carries 180 of 523 pre-existing terse tests without docstrings (65.6%), all predating this PR. Every test this change *adds* has a docstring; retrofitting the legacy tests is out of scope for a test-hardening PR.

@coderabbitai — the docstring-coverage warning reflects pre-existing tests, not this change; for your knowledge base, this repo's older `test_*.py` functions are intentionally terse, leaning on self-documenting `test_<behavior>_is_<code>` names rather than docstrings.

aallan:
@coderabbitai — verified your outside-diff finding on `tests/test_checker.py` (the positional `errs[0].error_code` assertions in the with-clause / handler tests): **already resolved in `715cdfd`**. All ten positional assertions in that family were converted to the order-independent `any(e.error_code == "..." for e in errs)` pattern; `grep -c 'errs\[0\]\.error_code' tests/test_checker.py` returns 0 on the current head. The pr-review-toolkit flagged the same fragility independently — it reproduced a `1 failed` full-suite run under one ordering — so this is fixed and validated. Thanks for the catch.

aallan:
@coderabbitai Re your second outside-diff finding (`tests/test_obligations.py:1053`, the generic smoke test) — tightened `assert result.ok` → `assert result.ok is True` in `81fd41d`, matching the `is False` identity-checks elsewhere in the file so it also catches an `ok=None` / truthy-but-not-`True` mutation. Thanks — and good catch on the consistency.

Issue #570, opened 2026-05-05
GC shadow-stack overflow in array_map (and sibling iterative builders) at ~4000 heap-allocating elements


aallan:
## Summary

`array_map(arr, fn)` over a heap-allocating closure body (e.g. `MkBox(@Int.0)`) traps on GC shadow-stack overflow at around 4000 elements. Each iteration's `gc_shadow_push` consumes a slot on the 16 KiB / 4096-entry shadow stack and the per-iteration push is not unwound within the loop, so allocations across many iterations leak shadow-stack slots until the limit is hit.

Surfaced during work on [#348](https://github.com/aallan/vera/issues/348) (GC worklist overflow); the natural runtime regression test for #348 (build a wide `Array<ADT>` and verify mark-phase correctness) hits this issue first.

## Reproducer

```vera
private data Box { MkBox(Int) }

public fn main(@Unit -> @Int)
  requires(true) ensures(true) effects(pure)
{
  let @Array<Box> = array_map(
    array_range(0, 5000),
    fn(@Int -> @Box) effects(pure) { MkBox(@Int.0) }
  );
  array_fold(
    @Array<Box>.0,
    0,
    fn(@Int, @Box -> @Int) effects(pure) {
      @Int.0 + match @Box.0 { MkBox(@Int) -> @Int.0 }
    }
  )
}
```

```
$ vera run /tmp/repro.vera
Error: Reached `unreachable` WASM instruction
Source backtrace:
  in anon_0  (line 7)   ;; the inner MkBox closure
  in main
```

Threshold (verified empirically): N=1000 ✓, N=3000 ✓, N=5000 ✗.

## Root cause

The lifted closure body for `MkBox(@Int.0)`:

1. Saves `$gc_sp` to `gc_sp_save` (per-call prologue)
2. Allocates the Box via `$alloc`; the new ptr is `gc_shadow_push`-ed
3. Returns; the epilogue restores `$gc_sp` from `gc_sp_save`, unwinding the push

That per-call balance is correct — the bug is upstream. `array_map`'s loop in the host runtime calls the lifted closure once per element, and the *outer* lifted-function frame for `array_map`'s body keeps each newly-allocated Box rooted on the shadow stack to survive subsequent `$alloc` calls in the same iteration. Those roots stack up across iterations because `array_map` doesn't have a "release per-iteration roots after the element is stored into the destination array" hook.

[...]

Issue #406, opened 2026-03-27
WASI 0.3 compliance and native async I/O


aallan:
WASI 0.3 adds native async I/O via `future<T>` and `stream<T>` types (stable in wasmtime 37+, November 2025). This is the prerequisite for true concurrent request handling in Vera's server-side effects.

Depends on WASI 0.2 compliance (#237).

## Scope

- Migrate from the WASI 0.2 incoming-handler model to WASI 0.3 async I/O
- Wire native `future<T>`/`stream<T>` to Vera's `<Async>` effect (currently eager/sequential)
- Enable concurrent request handling in `<HttpServer>` (#305) and `<McpServer>` (#306)
- Support WebSocket/SSE streaming (#228)

aallan:
Marking as **blocked by #237** via native GitHub issue relationship (now visible in the Relationships panel above).

WASI 0.3 builds on WASI 0.2 — the 0.2 compliance audit is a prerequisite to addressing 0.3 features. Already noted in the body and `ROADMAP.md` Phase 4a (`Depends on #237`); now also expressed structurally.

aallan:
Cross-link: #770 (non-blocking / timed stdin + input concurrency) names WASI 0.3 async I/O as the principled long-term substrate for the input half — a non-blocking stdin **stream** via `future<T>` / `stream<T>` is what would let a read be raced against a timer without blocking. (#770 also notes a near-term host-import timed read could land before the full async substrate.)

aallan:
**Gate recorded (sprint planning, 2026-07-02).** This issue is sequenced last in the server-effects sprint, behind an explicit gate:

- WASI 0.3.0 released 2026-06-11; wasmtime 46 ships it with Component Model Async enabled by default.
- **wasmtime-py 45.0.0 (verified against the installed wheel): component calls are sync-only.** `component.Func.__call__` marshals through `wasmtime_component_func_call`; `wasmtime_component_func_call_async` and the async config flags exist at the raw FFI layer but have zero Python plumbing. There is no implementation path for native component async from Python today.

**The gate:** wasmtime-py exposes component async at the Python API level (likely 46+, possibly needing an upstream contribution). Re-checked at each sprint release.

**Fallback ladder if the gate stays closed:**
1. Ship a "WASI 0.3 interface retarget with blocking lowering" release — the emitted world imports the 0.3 interface revisions, async-lowered imports deferred.
2. Re-title this issue as research/tracking with the gate as an explicit precondition and take it out of the sprint.

[...]

PR #845, opened 2026-07-02
feat(effects): <HttpServer> — verified HTTP request handling via vera serve — release v0.0.193

### LINKED ISSUE #305: <HttpServer> effect — verified HTTP request handling
## Summary

Add an `<HttpServer>` algebraic effect that allows Vera programs to handle incoming HTTP requests with verified contracts.

## Design

Vera's algebraic effect system maps naturally to HTTP server patterns. An incoming request is an effect operation; the handler provides the routing and response logic:

```
effect HttpServer {
  op on_request(HttpRequest -> HttpResponse);
}

handle[HttpServer] {
  on_request(@HttpRequest) -> {
    // route, process, build response
    resume(response)
  }
} in {
  serve(8080)
}
```

Contracts on handler operations can verify:
- Response status codes match expected patterns
- Required headers are present
- Response body conforms to a schema

## Dependencies

- #57 — HTTP client (shared `HttpRequest`/`HttpResponse` types)
- #237 — WASI 0.2 compliance (`wasi:http/incoming-handler`)
- #238 — Component Model (WIT interop for host bindings)

## Implementation notes

- Maps to WASI `wasi:http/incoming-handler` interface
- The `handle[HttpServer]` pattern models request routing as an effect handler
- Long-running server loop requires the `Diverge` effect (already supported)
- Real concurrency requires WASI 0.3 / `handle[Async]` (#270)

aallan:
## What

**Verified HTTP request handling** — the `<HttpServer>` effect (#305) — and **release v0.0.193**. Stage B of the server-effects sprint (`WASI.md`).

A server program defines a **total, contract-checked** handler and serves it:

```vera
public fn handle(@Request -> @Response)
  requires(true) ensures(true) effects(<HttpServer>)
{ ... }
```

```bash
vera serve examples/http_server.vera --port 8080
```

The accept loop lives in the **host** — handlers need no `Diverge`, are termination-checked like any function, and every contract on them is an ordinary Tier-1/Tier-3 obligation. The new `examples/http_server.vera` proves its status-range postcondition (`result == 200 || result == 404`) **statically at Tier 1** — the headline: verified HTTP handlers.

## Design

- **`HttpServer` is a marker effect** (no operations; spec §7.7.5), registered like `Async`/`Diverge`. `Request` (method, path, headers, body) and `Response` (status, headers, body) are **prelude ADTs**, injected only when referenced (the Json/HtmlNode conditional pattern — an always-inject first cut broke six no-overhead WAT pins because the `Map` headers field pulls heap machinery into pure programs). User definitions shadow them.
- **Instance-per-request isolation**: each request runs a fresh `execute()` (Stage-0 spike measured instantiation at ~0.02 ms). `State<T>` cannot leak between requests — pinned by test. Sequential handling in v1 (concurrency: #406).
- **Layout-driven marshalling, never hardcoded**: `CompileResult.adt_layouts` exports the constructor layouts this compilation computed; `build_request_adt` / `decode_response_adt` (vera/runtime/heap.py) fail loudly on an unexpected layout shape. A new `InstanceCaller` adapts a (Store, Instance) pair to the exact caller protocol the heap helpers already use (export lookup + store-context delegation) — validated empirically before building on it.
- **GC discipline**: every intermediate allocation in the Request build is shadow-rooted (#570/#692 class); an eager-GC round-trip test pins it, and dropping the pushes under mutation flips that test RED.
- **Contract violation → 500, always answered**: a runtime trap inside a handler maps to a 500 whose JSON body carries the trap diagnostic (`trap_kind`, message, frames — the `vera run --json` envelope shape).
- **Fixed en route**: Pass-1 fn signatures referencing prelude ADTs in params/return were computed before prelude registration and recorded `unsupported` in `fn_param_types` (the fn still compiled in Pass 2); a post-injection pass now re-registers exactly those (`_register_fn` is overwrite-idempotent). Functions may also now *declare* prelude-ADT-typed signatures servably.
- **Native-only**: the browser runtime does not serve HTTP (documented divergence, spec §12.9.3).

## Tests (all RED-first; 4 mutations kill)

- `tests/test_serve.py` (8): GET/POST echo round-trips (method/path/headers/body all cross the boundary), status propagation, **contract-violation → 500 with `trap_kind`**, `State<Int>` isolation across requests, eager-GC round-trip, two clean `make_server` validation errors. All ephemeral ports.
- Checker battery (5): effect-row acceptance pin, registry introspection, prelude-type destructure/construct (`_check_clean` — the vacuous `_check_ok` variants were tightened when 3 of 4 passed pre-change), State composition, status contract.
- Codegen compilability/export test; CLI validation tests (missing file / type error before bind / missing-handler message).
- Mutations: layout-shape guard broken → e2e RED; shadow pushes dropped → eager-GC RED; status offset shifted → status test RED; (plus the tier-count corpus pin moving 271/92/363 → 277/92/369 exactly by the example's 6 Tier-1 obligations).

## Conformance + example

- `tests/conformance/ch09_http_server.vera` at **level `verify`** (needs no network, unlike `ch09_http`/`ch09_inference` at `check` — rationale in the program header): **104 programs**.
- `examples/http_server.vera` (36 examples), `examples/README.md` row.

## Docs

Spec §7.7.5 + §9.5.6 + §12.9.3; SKILL.md HttpServer section + effect list; TOOLCHAIN.md serve recipe; CLAUDE.md commands; TESTING/AGENTS/FAQ counts (suite at 5,615 via the corpus-parametrized tests); allowlist refresh with three fixer duplicate-key collisions resolved against the real spec lines.

## Release v0.0.193

Version sync (6 sites), CHANGELOG cut + link refs, ROADMAP milestone chain updated, HISTORY: Stage-16 row **plus** the "By the numbers" table gains a permanent **"HttpServer effect"** milestone column (per maintainer direction, parallel to the v0.0.101 Inference column), the trailing soundness column refreshed to v0.0.191, totals footer and the opening sentence brought current (1,800+ commits / 193 releases / 92 active days — from git data).

## Verification

- [x] full suite 5,615 + stress 26 green; mypy clean; `ruff check` (+ `--select S`) clean without `--fix`
- [x] conformance 104 / examples 36 / doc-counts / version-sync / site-assets / limitations / encoding / diagnostic-fields / spec+SKILL+README+FAQ+EXAMPLES example gates
- [x] `vera verify examples/http_server.vera` → 6/6 Tier 1
- [x] 4 mutation kills, restored green

Closes #305


aallan:
**Review round 1 (pr-review-toolkit code-reviewer) — recorded.**

One merge-blocking finding, fixed in the commit just pushed:

1. **Critical — the CI-only `check_spec_examples` gate failed on this branch** (it is not in pre-commit, which is why local hooks stayed green). Two genuinely new spec blocks needed FRAGMENT allowlist entries (the §7.7.5 `effect HttpServer {}` fence and the §9.5.6 `data Request/Response` fence), and the §9.5.6 insertion (+33 lines) had orphaned four pre-existing signature-fragment entries (`float_is_nan`, `url_join`, the `string_chars` family, `md_parse`) that `fix_allowlists.py` cannot re-anchor once detached. All six re-keyed to their fence lines, the three stale old keys removed; the gate now exits 0 with zero duplicate keys.

Everything else came back **verified clean**: the core.py TypeAliasDecl branch byte-intact vs main with the #305 re-registration correctly scoped; layout-driven marshalling with loud shape guards and full shadow-rooting; `http_response` confined to the http_request path; server validation/isolation/trap-to-500 correct; CLI flag parsing safe on trailing flags; release prep consistent across all six version sites and the By-the-numbers metrics verified against the filesystem.

aallan:
@coderabbitai review

Issue #692, opened 2026-05-22
html_parse and json_parse trap with Out-of-bounds memory access on inputs that pressure GC during host-side tree walk


aallan:
## Summary

**Both `html_parse` and `json_parse` trap with `Out-of-bounds memory access`** on inputs large enough to pressure GC during host-side tree marshalling, escaping their declared `Result<HtmlNode, String>` / `Result<Json, String>` contracts.  Confirmed root cause: **missing shadow-stack rooting in `vera.wasm.html_serde.write_html` and `vera.wasm.json_serde.write_json`** — the same bug class as #570 (`array_map`), #515 (fold accumulator), and #593 (closure-return shadow asymmetry), but on the **host side** rather than WAT-emitted user code.

External report from an agent building a `summarise_urls` example surfaced the `html_parse` trap.  The agent's bisection ruled out a pure size threshold — same byte count triggers or doesn't depending on what's at both ends of the window — which is exactly what GC-pressure-driven corruption looks like (it depends on what got allocated *before* the host walker ran).

The parallel write_json bug is **proven, not claimed** — reproducer and trap output below.

### `html_parse` (original report)

```vera
effect IO { op print(String -> Unit); }

public fn main(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
  let @Result<String, String> = IO.read_file("/path/to/FAQ.md");
  match @Result<String, String>.0 {
    Ok(@String) -> {
      let @Result<HtmlNode, String> = html_parse(@String.0);
      match @Result<HtmlNode, String>.0 {
        Ok(@HtmlNode) -> IO.print("Ok\n"),
        Err(@String) -> IO.print("Err\n")
      }
    },
    Err(@String) -> IO.print("read failed\n")
  }
}
```

Output:
```
Error: Out-of-bounds memory access
Source backtrace:
  (suppressed 1 runtime-helper frame above first user code)
  in main  (file.vera:8-22)
```

Neither `Ok` nor `Err` is printed — the trap bypasses both match arms.

### `json_parse` (verified during triage)

Generate a deep nested array (1000 sub-arrays × 50 numbers, ~192 KB input):

```python
import json
data = [list(range(50)) for _ in range(1000)]
text = json.dumps(data)
# embed `text` into a Vera string literal that calls json_parse(@String.0)
```

```vera
effect IO { op print(String -> Unit); }

public fn main(@Unit -> @Unit)
  requires(true) ensures(true) effects(<IO>)
{
[...]

PR #859, opened 2026-07-02
fix(runtime-gc): ShadowGuard.push bounds the full 4-byte slot and rejects negative sp

### LINKED ISSUE #791: _ShadowGuard.push should reject a partial shadow-stack slot (sp + 4 > limit)
Surfaced by CodeRabbit's review of the #421 decomposition (#789). **Pre-existing** — moved verbatim from `codegen/api.py`. Defensive / theoretical (low priority).

`vera/runtime/heap.py`'s `_ShadowGuard.push` guards overflow with `if sp >= limit:` before writing 4 bytes at `[sp..sp+3]`. If `gc_sp` were ever left with only 1–3 bytes of headroom (`sp == limit - 1/2/3`), the check passes and `push` writes past the shadow-stack window instead of raising the intended overflow diagnostic. The WAT side always advances `gc_sp` in 4-byte steps and 4-aligns the window, so this can't occur in practice today — but a stricter `if sp < 0 or sp + 4 > limit:` is a cheap robustness improvement against a future hand-written fixture or host path.

aallan:
## Summary

Burndown PR for #791. `_ShadowGuard.push` guarded shadow-stack overflow with `sp >= limit` before writing a 4-byte slot, so a misaligned `gc_sp` with 1–3 bytes of headroom would write past the window; a negative `sp` would index the ctypes buffer before the linear-memory base. The bound is now slot-complete: `sp < 0 or sp + 4 > limit`.

**Honest framing** (per the issue and the inventory): the misaligned state is unreachable today — generated WAT always advances `gc_sp` in 4-byte aligned steps — so this is defense-in-depth hardening of the GC trust root, not a live exploit. That said, the orchestrator's independent flip showed the negative-`sp` case **hard-crashes the host interpreter** on the unfixed tree (out-of-bounds ctypes read), which is exactly the class of failure a trust-root guard exists to make impossible.

- New `TestShadowGuardPushBound791` battery (6 tests) constructs the misaligned state directly at the unit level (hand-rolled WAT exporting the #692 triple, `gc_sp = limit − 1/2/3`): 3 clean REDs (`DID NOT RAISE`) on the unfixed tree, plus boundary pins (exact-final-slot accepts, full-window rejects) green on both sides as controls.
- Mutation-validated: bound reverted → 3 RED. Sibling sweep: `heap.py` has no other off-by-slot bound (`_read_string_export` already slot-complete). Four structurally-identical bounds exist in the WAT emitters and browser runtime, each protected by the same alignment invariant — tracked as a follow-up hardening issue, out of this PR's scope.
- Gates: 5,793 passed / 43 skipped; mypy, ruff, 106 conformance, 37 examples, GC battery 38/38, eager-GC probes clean.

Closes #791

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened shadow-stack bounds handling so writes now fail safely unless a full 4-byte slot is available, including edge cases with negative offsets.
  * Added extra protection against out-of-bounds memory writes in host-side GC rooting scenarios.

* **Tests**
  * Added regression coverage for boundary conditions around shadow-stack pushes, including exact-limit success and invalid-offset rejection.

* **Documentation**
  * Updated changelog, testing notes, and roadmap figures to reflect the latest coverage and hardening work.


aallan:
Adversarial review round 1 (single-skeptic, S-mechanical tier, lens: does the RED battery pin the flaw): CLEAN, no findings at any severity.

Attack evidence: reverting the bound flips the three partial-slot tests RED (`DID NOT RAISE`); an over-tight mutation (`sp + 4 >= limit`) is caught by the exact-final-slot accept test, so both sides of the boundary are pinned; a write-before-raise probe confirmed the no-partial-write memory assertion catches real corruption (`b'4\x12\x00\x00'` detected at the window edge); a full caller sweep confirmed `gc_sp` is only ever 4-aligned in reachable states, so the new bound changes no reachable behavior; CHANGELOG structure verified intact. Orchestrator's independent flip additionally showed the negative-`sp` case hard-crashes the interpreter on the unfixed tree (out-of-bounds ctypes read).

aallan:
@coderabbitai review

Issue #798, opened 2026-06-25
verifier/smt.py: integer arithmetic overflow unmodeled — Tier 1 proves contracts false under i64/u64 wraparound


aallan:
Found by the #392 `smt.py` soundness audit. Parent: #392. Sibling of #797 — same root cause: idealized Z3 sorts modeling Vera's machine types.

## The bug
`vera verify` proves a postcondition the runtime then rejects — a direct Tier-1 / Tier-3 disagreement, for both `@Int` and `@Nat`:
```
public fn inc(@Int -> @Int)            # identical shape for @Nat at the u64 ceiling
  requires(true) ensures(@Int.result > @Int.0) effects(pure)
{ @Int.0 + 1 }
```
`vera verify` → `verification: {tier1_verified: 2}`, zero diagnostics — `result > input` is **proved for all inputs**. But the always-on runtime postcondition check traps:
```
$ vera run inc -- 9223372036854775807     # MAX_i64
Error: Postcondition violation in inc(@Int -> @Int)
  ensures(@Int.result > @Int.0) failed
```
At `x = MAX_i64`, `x + 1` wraps to `-9223372036854775808` (`< x`), so the *proved* postcondition is false at runtime. `vera verify` says ✓; the verified program traps.

## Root cause
`declare_int` (`vera/smt.py:198`) → `z3.Int` (unbounded ℤ); `declare_nat` (`:210`) → `z3.Int ∧ ≥0` (unbounded ℕ). No overflow modeling anywhere in the verifier / smt / obligations layer (grep-confirmed). Vera `@Int` / `@Nat` are wrapping `i64` / `u64`, so any Tier-1 proof implicitly relying on no-wraparound is unsound.

## Inconsistency with the existing partial-arithmetic story
Vera already guards two partial-arithmetic cases at runtime *and* assumes-the-safe-case at Tier 1:
- `@Nat` underflow → `nat_sub` obligations + Tier-3 trap (spec §4, §11).
- Signed-division `MIN / -1` overflow → `i64.div_s` trap (spec §12).

But add / sub / mul overflow wraps silently — neither trapped at runtime nor obligated at Tier 1. Overflow is the one partial-arithmetic case treated as total by *both* tiers.

## Fix options (a language-semantics decision)
[...]

Issue #758, opened 2026-06-18
Verifier: `@Nat` narrowing not obligated at function return position or value-position tuple/constructor components


aallan:
Surfaced by the independent multi-agent review of PR #756 (part of the #747 narrowing-obligation work). Two `@Nat` narrowing **positions** are not statically obligated by the verifier, so a `vera verify`-clean program can store/return a negative `@Nat`. Both are **pre-existing** — reproduced on v0.0.172 (base `7fb53ac`), *not* introduced by #747/#756 — but #747 generalized the binding-site obligation and its docs implied completeness, so these are the remaining holes. Documentation softened in #756 to stop claiming universal soundness and to point here.

## 1. Function return position (confirmed firsthand)

```vera
public fn to_nat(@Int -> @Nat) requires(true) ensures(true) effects(pure) { @Int.0 }
```

`vera verify` → `OK: 4 verified (Tier 1)`, no E503/E504. At runtime `to_nat(0 - 5)` returns **`-5`** through the `@Nat` slot, no trap. The narrowing walker `_walk_for_nat_binding_obligations` (`vera/verifier.py`) walks the body for binding *sites* but never treats the function's own return slot as a narrowing site, and codegen emits no return-coercion guard. The implicit `@Nat` result invariant is enforced only if the author writes `ensures(@Nat.result >= 0)` explicitly.

## 2. Value-position tuple / constructor components (confirmed firsthand)

```vera
public fn mk(@Int -> @Tuple<Nat, Nat>) requires(true) ensures(true) effects(pure) { Tuple(@Int.0, 5) }
```

`vera verify` → `OK: 2 verified (Tier 1)`, no obligation on the `@Int.0 -> @Nat` first-component narrowing. The per-component `@Nat` target of a `Tuple(...)` / constructor call built in value (return / argument) position is not threaded down to the call's arguments, so the field-narrowing loop never fires. (Contrast the `let`-destructure path, which *is* obligated.)

## Fix sketch

Both are statically checkable today via the existing checker→verifier side-table (`expr_semantic_types` / `expr_target_types`):

- **Return slot:** in the narrowing walker, treat the function's declared return type vs. the body tail-expression's result type as a narrowing site (mirror the existing binding-site obligation + the Tier-3 runtime guard).
[...]

Issue #520, opened 2026-04-23
Nat subtraction silently underflows to negative i64 — refinement-type soundness hole


aallan:
## Summary

`@Nat` subtraction silently underflows to a negative i64. The type system accepts `@Nat - @Nat : @Nat` and the codegen emits a plain `i64.sub` — no runtime check, no promotion to `@Int`. A `@Nat` local can hold a negative value at runtime despite the type's claim that it represents non-negative integers. This is a **soundness issue** with the refinement type system, not just a documentation gap.

## Reproducer

```vera
public fn test_underflow_trap(@Nat -> @Nat)
  requires(true) ensures(true) effects(pure)
{
  @Nat.0 - 1
}

public fn main(@Unit -> @Int)
  requires(true) ensures(true) effects(pure)
{
  -- Pass 0 to underflow-prone subtraction; type system allows it,
  -- runtime returns -1 bound to a @Nat local.
  let @Nat = test_underflow_trap(0);
  nat_to_int(@Nat.0)
}
```

`vera check` ✅
`vera verify` ✅ (the `requires(true)` ensures no contract objection)
`vera compile` ✅
`vera run` prints `-1`.

The `@Nat` local holds a negative value. Any downstream code that relied on the `Nat >= 0` invariant to reason about correctness now has a silent bug.

## Why this matters

Vera's core selling point is contract-verified code with refinement types that the compiler checks at the border between user code and helpers. When a type system promises "this value is non-negative" and the runtime produces a negative value in that slot, every contract that transitively depends on that invariant is undermined — including contracts the verifier proved sound under the (now violated) assumption.

Concretely: an `ensures(@Nat.result < array_length(arr))` contract on a helper that internally does `@Nat.0 - k` with an insufficient precondition can pass the verifier (Tier 1), compile, and still produce an out-of-bounds access at runtime via the negative-`@Nat` path. The `Array.[...]` indexing operator trusts its `@Nat` index not to be negative; a negative `@Nat` therefore becomes a memory-safety concern on top of a logical correctness one.

## Discussion: which semantics should Vera adopt?

Four reasonable choices, each with precedent:

1. **Runtime trap on underflow** (Rust `u64::checked_sub` + panic, or `.unwrap()` default). Straightforward; minimal compile-time disruption; programs crash loudly on the bug instead of silently. Adds one compare-and-branch per `@Nat` subtraction.

[...]

PR #862, opened 2026-07-02
fix(checker): Int op Nat types as Int — the formal LUB, not a silent Nat assertion

### LINKED ISSUE #755: Checker types `Int <op> Nat` as `Nat`, inferring spurious @Nat narrowings
Surfaced while implementing #747 (the @Nat binding-site narrowing generalization).

## Finding

A non-negative integer literal is typed `Nat`, and the binary-arithmetic rule promotes the result to `Nat` whenever **either** operand is `Nat` — so an `Int <op> Nat-literal` expression is typed `Nat` even though it can be negative:

```
@Int.0 + 2   ->  Nat
@Int.0 * 2   ->  Nat
@Int.0 - 2   ->  Nat     -- clearly unsound: Int - 2 can be negative
```

(reproduced via `typecheck_with_artifacts` on `fn f(@Int -> @Int) { @Int.0 <op> 2 }`).

## Impact

The `Int - Nat` case is masked downstream by #520's underflow-leaf check and #552's narrowing obligation (a pure-literal subtraction is still obligated `>= 0`), so it does not produce unsound *verified* code today. But the promotion drives **spurious @Nat narrowing inferences** at other sites: an `async(@Int.0 * 2)` infers `Future<Nat>`, so the `@Int -> @Nat` narrowing fires (counterexample `@Int.0 = -1`) on an expression the author never annotated as `Nat`. `tests/conformance/ch09_async.vera` was worked around with an explicit `requires(@Int.0 >= 0)` on its helper rather than letting the inferred `Nat` stand.

## Fix direction

`Int <op> Nat` (and `Nat <op> Int`) should type as `Int` for addition/multiplication/subtraction — the result is only `Nat` when **both** operands carry `Nat` provenance. The non-negative-literal-is-`Nat` rule is fine in isolation; the unsoundness is in the binop promotion. See the numeric-type join logic in `vera/checker/` and `vera/types.py`.

aallan:
## Summary

Burndown PR for #755. Mixed `Int <op> Nat` arithmetic typed as `Nat` whenever the Int operand came first — the join used bidirectional `is_subtype`, and the verifier-mediated narrowing relaxation (`Int <: Nat`, spec §2.8 rule 5) leaked into the checker's widening join. Typing `@Int.0 - 2` as `Nat` silently asserts non-negativity with no obligation behind it (spurious `Future<Nat>` inferences, E503s the author never asked for), violating §0.2.2 no-implicit-behaviour.

- New `numeric_join` LUB helper in `vera/types.py`: only `Nat <: Int` (the formal refinement subtyping, §2.2.1) participates; `Int op Nat` → `Int` for all five operators, both operand orders. `Nat op Nat` → `Nat` and `Int/Float64` mismatches (E141) unchanged.
- Spec §4.4 gains the explicit mixed-arithmetic result-type rule; the checker now agrees with what the #798 overflow classifier and #520 subtraction guard already assumed.
- Blast radius handled honestly: one overflow-differential test re-baselined (its load-bearing i64-width assertion preserved, plus a new mixed-operand case restoring lost coverage); `ch09_async.vera`'s `requires(@Int.0 >= 0)` workaround for this exact bug removed.
- Obligations shift as expected: `@Int.0 - 2` now carries `int_overflow` (was `nat_sub`); genuine `@Nat - @Nat` keeps its underflow guard (verified).
- Tests observe the synthesized type directly via `typecheck_with_artifacts` (cannot coincide with a fallback); 3 RED on the unfixed tree; mutation kill: reverting the join flips them RED.
- Residual note: the runtime negative-Nat at a bare return-position narrowing is pre-existing and already tracked by #758 (spec §11.2.1 documents it) — this PR makes the *type* honest, #758 owns the guard.

Orchestrator flip: the branch's type-observation tests fail 3/6 against main's compiler, 6/6 pass on the branch.

Closes #755

## Summary by CodeRabbit

* **Bug Fixes**
  * Mixed `Int`/`Nat` binary arithmetic now consistently yields an `Int` result type (with `Nat <op> Nat` staying `Nat` and `Int <op> Int` staying `Int`).
  * Overflow/narrowing behaviour for mixed numeric expressions into `@Nat` contexts is corrected to match the intended typing rules.

* **Tests**
  * Added regression coverage for the mixed `Int`/`Nat` result-typing cases, plus updated overflow-related expectations and metrics.

* **Documentation**
  * Specification updated to clarify the least-upper-bound result typing rule for mixed arithmetic, and related project documentation refreshed.


aallan:
@coderabbitai review

aallan:
@coderabbitai review

Issue #861, opened 2026-07-02
Refinement-alias predicates skip type-checking: non-Bool predicates and E142-class expressions pass vera check


aallan:
## Summary

The predicate inside a refinement type alias is not type-checked the way contract predicates are:

1. `type T = { @Byte | @Byte.0 }` — a **non-Bool** predicate (bare `@Byte` value) — passes `vera check`.
2. `@Byte.0 < 10` is accepted inside a refinement predicate, while the identical expression in a `requires` clause is rejected with E142 (Byte has no ordered comparison with an int literal at the checker level).

`vera/checker/core.py`'s `_check_contract` has no refinement counterpart — the alias predicate appears to skip well-formedness checking entirely. The codegen consequence of one such unchecked shape was #766 (i64/i32 guard-width trap), fixed separately; this issue is the checker-phase laxness that let it through.

## Expected

Refinement predicates get the same checking as contract predicates: must type as `Bool`, and operand typing rules match the rest of the checker (whatever E142's rule decides for `@Byte` comparisons — the checker, guard codegen, and spec §2.6 must agree; DESIGN.md decides any ambiguity).

## Notes

Found during the #766 fix's sibling-width probe (2026-07-02 burndown); joins the v0.1.0 bug queue. KNOWN_ISSUES Bugs row rides the #766 fix PR as the discovering change.

Issue #773, opened 2026-06-21
Eq auto-derivation is scalar-rep-based, not structural (false-rejects String-field ADTs, false-accepts ADT-pointer fields)


aallan:
## Summary

Codegen's `Eq` auto-derivation is **scalar-rep-based**, not structural. `_adt_satisfies_eq` accepts a field only if its WASM rep is scalar (i64/i32/f64), and `_translate_adt_eq` compares each field with a scalar `.eq`. Measured against the documented guideline (`Eq` is satisfied by primitives — including `String` — and ADTs whose fields are themselves `Eq`), this is wrong in **both** directions:

- **False-reject**: a `String` field is `i32_pair`, so a `String`-field ADT (e.g. `Box<String>`) is rejected with `E613` even though `String` satisfies `Eq`.
- **False-accept**: a concrete *ADT* field is an `i32` pointer, so it passes the scalar check and is compared with `i32.eq` — **pointer identity, not value**. Structurally-equal values with distinct heap pointers compare unequal (and aliased pointers compare equal regardless of value).

## Repro (false-reject)

```vera
public data Box<T> { MkBox(T) }
private forall<T where Eq<T>> fn eq2(@T, @T -> @Bool)
  requires(true) ensures(true) effects(pure) { @T.1 == @T.0 }
public fn main(@Unit -> @Bool) requires(true) ensures(true) effects(pure)
{ let @Box<String> = MkBox("a"); eq2(@Box<String>.0, @Box<String>.0) }
```

`vera compile` rejects this with `E613` though `String` is `Eq`. The false-accept is the dual: an ADT with a concrete nested-ADT field compiles and compares that field by pointer.

## Root cause

The `Eq` check (`vera/codegen/monomorphize.py` `_adt_satisfies_eq`) and the eq codegen (`vera/wasm/operators.py` `_translate_adt_eq`) both work from WASM reps, not Vera field types. `_translate_adt_eq` expands the comparison **inline** and has no `String`-field path and no nested-ADT recursion.

## Proper fix (a feature)

Structural `Eq` derivation:
- Thread Vera field types into the constructor layout (it stores only `(offset, wasm_type)` today).
- Generate per-ADT **recursive** `$eq_<ADT>` WASM functions (inline expansion can't handle recursive types like `List<T>`), dispatching per field type: scalar `.eq`, `String` byte comparison, nested-ADT recursion, `Array`/handle → not derivable.
[...]

PR #864, opened 2026-07-02
fix(codegen): @Byte refinement predicates lower at i32 with the 0..255 range conjoined

### LINKED ISSUE #766: Refinement guard over `@Byte` emits an i64 comparison against the i32 value → invalid WASM at run time
## Summary

A refinement over `@Byte` (`type SmallByte = { @Byte | @Byte.0 < 10 }`) passes `vera check`, `vera verify`, and `vera compile`, but **`vera run` fails to instantiate the module** — wasmtime rejects the generated WASM with a translation error. A *plain* `@Byte` parameter compiles and runs fine, so the defect is specific to the **refinement runtime guard** on a Byte value.

```
$ vera run smallbyte.vera --fn f -- 5
Error: Unhandled WASM trap: failed to compile: wasm[0]::function[1]::f
Caused by:
    0: WebAssembly translation error
```

## Root cause

`@Byte` is represented as **i32** at runtime, but the refinement-predicate runtime guard lowers `@Byte.0 < 10` as an **i64** comparison (the predicate/SMT layer models the base as an integer → i64). The emitted guard reads the i32 Byte parameter and compares it with `i64.lt_s`, an i32/i64 stack-type mismatch that the WASM validator rejects:

```wat
(func $f (export "f") (param $p0 i32) (result i64)   ;; $p0 is i32 (Byte)
  local.get 0          ;; i32
  i64.const 10         ;; i64
  i64.lt_s             ;; <-- compares i32 operand with i64.lt_s: INVALID
  i32.eqz
  if
    i32.const 0
    i32.const 78
    call $vera.contract_fail
    unreachable
  end
  i64.const 0)
```

## Reproduction

```vera
type SmallByte = { @Byte | @Byte.0 < 10 };
public fn f(@SmallByte -> @Int)
{ 0 }
```

`vera check`/`verify`/`compile` succeed; `vera run --fn f -- 5` fails as above. A plain `public fn g(@Byte -> @Int) { 0 }` runs fine.

It's a loud failure (the module won't instantiate), not a silent wrong answer — but a check/verify/compile-clean program shouldn't fail to run.

## Suggested fix

In the Byte refinement guard codegen, reconcile the value's i32 representation with the i64-lowered predicate — either widen the i32 Byte to i64 (`i64.extend_i32_s`/`_u`) before the comparison, or lower a `@Byte` predicate with i32 ops. While there, conjoin the implicit Byte range invariant (`0 <= @Byte.0 <= 255`) the way the `@Nat` guard conjoins `>= 0` (the original PR #763 review point — deferred because the guard didn't compile at all).

## Relationship

Surfaced during the #746 review (PR #763), where the range-conjoin suggestion was declined because the guard doesn't compile. Runtime-guard-codegen gap in the same family as the deferrals tracked in #764 / #765.

aallan:
## Summary

Burndown PR for #766. A refinement over `@Byte` (`{ @Byte | @Byte.0 < 10 }`) passed check/verify/compile but trapped at instantiation (`type mismatch: expected i64, found i32` at offset 79): `@Byte` is i32 (spec §11), but the runtime guard lowered predicates at i64 — int literals as `i64.const`, comparison/arithmetic from the default i64 op tables. Only refinement predicates reach this (ordinary `@Byte < 10` is E142 at the checker), and the verifier correctly defers Byte refinements to Tier 3 — so the runtime guard was the only check, and it was invalid WASM. A broken documented feature.

- The sibling-width probe showed the divergence was **not comparison-only**: arithmetic (`@Byte.0 + 1`), equality, logical, and all comparison shapes hit the same two emission sites. The fix dispatches any binop with a `@Byte` operand entirely at i32 (unsigned cmp + div/rem, literals coerced to `i32.const`) in `vera/wasm/operators.py` — `_translate_adt_eq` untouched, staying disjoint from #773's territory.
- `vera/codegen/contracts.py` conjoins the implicit `0 <= @Byte.0 <= 255` range, mirroring the `@Nat >= 0` conjoin — an FFI Byte crossing as unbounded i32 (e.g. 300) is now rejected, not laundered.
- 5 RED execution tests (exact offset-79 signature on the unfixed tree) + new conformance program `ch02_byte_refinement` (suite now 107). Mutation kills isolate cleanly: width dispatch off → 5 RED; range conjoin off → only the range test RED.
- Discovered during the sibling probe and filed as #861 (row added here as the discovering PR): refinement-alias predicates skip type-checking entirely — the checker-phase laxness that let this shape through. Fixed separately in the burndown queue.

Orchestrator flip: the branch's conformance program traps with the exact width signature under main's compiler and runs clean on the branch.

Closes #766


aallan:
@coderabbitai review

aallan:
@coderabbitai review

Issue #790, opened 2026-06-24
Math.log(0)/log2(0)/log10(0) return NaN instead of −∞ (Python/JS parity)


aallan:
Surfaced by CodeRabbit's review of the #421 decomposition (#789). **Pre-existing** — the wrapper moved verbatim from `codegen/api.py`.

`vera/runtime/math.py`'s `_math_unary_host` converts every `ValueError` from Python's `math` module into `NaN`, to mirror IEEE 754 / JS for genuine domain errors (e.g. `log(-1)` → NaN). But Python's `math.log(0.0)` raises `ValueError` (math domain error) at the **pole**, where IEEE 754 and JS `Math.log(0)` both give **−∞**. So Vera's `Math.log(0)` / `log2(0)` / `log10(0)` return NaN under the Python/wasmtime runtime while the browser/JS runtime returns −∞ — a parity violation.

**Fix:** special-case `x == 0` → `-inf` in `_math_unary_host` for `log`/`log2`/`log10`, keeping the `ValueError → NaN` path for true domain errors. Add a `tests/test_browser.py` parity case for `log(0)`.

PR #885, opened 2026-07-03
fix(codegen): float_to_string is total over non-finite floats — nan/inf/-inf, not a trap (#857)

### LINKED ISSUE #857: float_to_string traps on non-finite floats (NaN/±inf) with a raw or misleading diagnostic
## Summary

`float_to_string` is documented as total (SKILL.md: "returns String (decimal representation)") but traps on non-finite `Float64` values — which are legal, spec-sanctioned values (the math builtins commit to IEEE 754 semantics, so `log(0.0)` legitimately produces `-inf` and domain errors produce `NaN`).

Two distinct failure presentations, both wrong:

1. On current main, `IO.print(float_to_string(log(0.0)))` (NaN input) dies with a **raw WASM trap**: `wasm trap: invalid conversion to integer` — an `i64.trunc_f64` in the digit-extraction path, no E-code, no rationale.
2. With the #790 fix (`log(0.0)` → `-inf`), the same program dies with the **wrong diagnostic**: the E-coded `Integer overflow` trap message, whose Fix text tells the user to add integer `requires` preconditions — for a float-printing call. Diagnostics-as-instructions pointing in a wrong direction is exactly the failure class the design principles rule out (DESIGN.md; spec §0.2(1)).

## Expected

`float_to_string` total over all Float64 values: `NaN` → `"nan"`, `inf`/`-inf` → `"inf"`/`"-inf"` (exact strings to be chosen for cross-runtime parity — JS `String(-Infinity)` is `"-Infinity"`, so the browser runtime's current behavior must be confirmed and BOTH runtimes aligned on one spelling, spec'd in §9.6/§9.8).

## Repro

```vera
  requires(true)
  ensures(true)
  effects(<IO>)
{
  IO.print(float_to_string(log(0.0)))
}
```

## Notes

- Found 2026-07-02 during the #790 independent repro verification (the burndown's orchestrator flip check); pre-existing on main — not introduced by #790, but #790 makes `-inf` more reachable.
- Joins the v0.1.0 bug burndown queue; the KNOWN_ISSUES Bugs row rides the #790 fix PR as the discovering change.

aallan:
## Summary

`float_to_string` is documented total, but `NaN` died with a raw `wasm trap: invalid conversion to integer` (from `i64.trunc_f64_s` in the digit-extraction path) and `±inf` (−inf reachable from `log(0.0)` under the IEEE 754 semantics #790 shipped) surfaced a **misleading Integer-overflow diagnostic** steering the user toward integer `requires` preconditions on a float-printing call.

## Fix

A non-finite short-circuit in `_translate_float_to_string` (`vera/wasm/calls_strings.py`) *before* the trunc path: `f64.ne(x,x)`→`nan`, `f64.abs(x)==inf`→`inf`/`-inf` (sign via `f64.lt 0`), rendered via a new `_emit_write_ascii_literal` helper with a `done` flag gating the finite body. The finite path is byte-for-byte unchanged (carry case `1.9999995`→`1.999999`, large/tiny/negative all identical).

**Spelling + parity**: `nan`/`inf`/`-inf` (lowercase, matches Python `str()` and the issue's Expected). Parity is **by construction** — `float_to_string` compiles to inline WASM (no host import), so wasmtime and Node execute the same module bytes; a true node↔wasmtime differential (`test_non_finite_render_parity`) asserts byte-identical stdout, **passed not skipped**.

## Evidence

- **RED (main)**: NaN → `invalid conversion to integer`; −inf via `log(0.0)` and +inf → `Integer overflow` misdirection (verbatim).
- **Mutation kill**: neutering the short-circuit reverts to the NaN trap + ±inf overflow — flips the 5 tests RED.
- Full gate green: 5952 passed / 43 skipped; mypy + ruff (incl `--select S`) clean; 108 conformance; 37 examples.

Principle (DESIGN.md §0.2 (1)): loud and total over silent and partial; a diagnostic-as-instruction pointing the wrong way is the exact failure class ruled out.

Closes #857

## Summary by CodeRabbit

* **New Features**
  * `float_to_string` now converts every `Float64` to a string without failing, including `nan`, `inf`, and `-inf`.
  * `show(`@Float64`)` uses the same formatting for consistent results across runtimes.

* **Bug Fixes**
  * Fixed cases that previously trapped or produced misleading diagnostics for non-finite values, including the `log(0.0)` → `-inf` case.

* **Documentation**
  * Updated the standard-library specification and related descriptions to document the exact non-finite spellings and formatting rules.

* **Tests**
  * Added expanded non-finite rendering and Python-vs-browser parity coverage.


aallan:
@coderabbitai review

aallan:
@coderabbitai review

Issue #891, opened 2026-07-03
Generic instantiated over a parameterized ADT traps at run: 'expected i64, found i32' in the mono clone


aallan:
## Summary

A generic function instantiated over a **parameterized ADT** type (`gid(MkBox(7))` where `gid : forall<T> fn(@T -> @T)` and `data Box<T> { MkBox(T) }`) traps at `vera run` with `type mismatch: expected i64, found i32` inside the monomorphized clone `gid$Box` — a check/verify-green program that fails WASM compilation. The clone treats the type-variable slot `@T` as `i64` (the scalar default) while the concrete `Box` value is an `i32` heap pointer, so the emitted clone body is ill-typed.

This is NOT the #772 `Eq`-derivation false-accept (there is no `Eq` constraint here — `gid` is plain identity) and NOT the #769 type-inference-picks-the-wrong-type family (inference correctly resolves `T = Box`). It is a monomorphization **representation/boxing** gap: the clone's WASM signature and body use the scalar `i64` for a type variable bound to a pointer-represented ADT.

## Repro (purely local — no modules needed)

```vera
public data Box<T> { MkBox(T) }

public forall<T> fn gid(@T -> @T)
  requires(true) ensures(@T.result == @T.0) effects(pure) { @T.0 }

public fn main(@Unit -> @Box<Int>)
  requires(true) ensures(true) effects(pure) { gid(MkBox(7)) }
```

`vera check` / `vera verify` → OK; `vera run` → `Unhandled WASM trap: failed to compile ... gid$Box ... type mismatch: expected i64, found i32`.

Also reproduces with `Box<T>` defined in another module and imported (same trap in the same clone), so it is orthogonal to the cross-module work in #774 — the clone codegen is wrong regardless of where the ADT is declared.

## Root cause (sketch)

`Monomorphizer.monomorphize_fn` substitutes `T -> Box` and the clone's parameter/return WASM types are derived from the bare ADT name `Box`, which is a uniform `i32` pointer — but the type-variable slot lowering in the clone body (and/or the clone's signature) still assumes the scalar `i64` default for `@T`. The clone must lower a type variable bound to an ADT as `i32` (a heap pointer), matching the ADT's representation, on both the signature and the body's slot reads.

## Pre-existing

[...]

Issue #772, opened 2026-06-21
Eq auto-derivation is type-arg-blind on the constructor-inferred path (false-accepts Box<String>)


aallan:
## Summary

A generic constrained `forall<T where Eq<T>>` is accepted when called with a value whose type is **inferred from a constructor** — e.g. `eq2(MkBox("a"), …)` — even when the concrete type isn't `Eq`-derivable (here `Box<String>`, because `String` is an `i32_pair` field, not scalar-`Eq`). The generated equality then compares the wrong representation (the string pointer, not the value) at run time. This is a **false-accept** (a soundness gap), not a spurious rejection.

## Current state (post-#767)

#767 made `_adt_satisfies_eq` validate type arguments — but only on the **parameterized-name** path:

- A *slot reference* carries the full type: `let @Box<String> = MkBox("a"); eq2(@Box<String>.0, …)` passes `Box<String>` to the `Eq` check, which splits base + args and **correctly rejects** it (`E613`). `Box<Int>` is accepted.
- A *constructor* inference resolves to the **bare** ADT name: `eq2(MkBox("a"), …)` infers the constrained type var as `Box` (no args), because the monomorphizer's `_infer_vera_type_name` returns `ctor_to_adt.get(name)` for a `ConstructorCall`. That bare name is correct for clone naming (the ADT is a uniform pointer, so one `g$Box` clone serves all `Box<T>`), but it leaves `_adt_satisfies_eq("Box")` with no type args to validate → it **false-accepts** `Box<String>`.

## Repro

```vera
public data Box<T> { MkBox(T) }

private forall<T where Eq<T>> fn eq2(@T, @T -> @Bool)
  requires(true) ensures(true) effects(pure) { @T.1 == @T.0 }

public fn main(@Unit -> @Bool)
  requires(true) ensures(true) effects(pure)
{ eq2(MkBox("a"), MkBox("a")) }   -- compiles; should be E613 (String is not Eq-derivable as a field)
```

The slot-ref form of the same program (`let @Box<String> = MkBox("a"); eq2(@Box<String>.0, @Box<String>.0)`) is correctly rejected after #767.

## Root cause

The ability check (`_check_constraints` → `_adt_satisfies_eq` in `vera/codegen/monomorphize.py`) receives the constrained type var's name from the monomorphizer's type-arg inference, which **drops type args** for a `ConstructorCall` (returns the bare ADT name). That's correct for cloning; lossy for the `Eq` check.

## Why this is deferred (the fix is broad)

[...]

PR #916, opened 2026-07-03
fix(codegen): composite `==` postcondition lowers to structural Eq, not a pointer compare (#912)

### LINKED ISSUE #912: SEVERE: composite-typed postcondition == is Tier-1 verified but traps at runtime (spurious postcondition violation)
## Summary (SEVERE — a verified program traps on its own proved postcondition)

A postcondition asserting **equality of a composite/ADT value** — `ensures(@T.result == <constructor>)` — is **verified Tier-1 by Z3** but **traps at runtime** with a spurious postcondition violation. The runtime postcondition check lowers composite `==` to raw `i32.eq` (pointer comparison) instead of structural equality, so a structurally-correct result fails the pointer compare and traps.

This directly undermines the core value proposition: a program that `vera verify` certifies as correct crashes at `vera run` on its own proved contract.

## Repro

```
private data Box { MkBox(Int) }
private fn mk(@Int -> @Box)
  requires(true) ensures(@Box.result == MkBox(@Int.0)) effects(pure)
{ MkBox(@Int.0) }
public fn main(@Unit -> @Int)
{ match mk(5) { MkBox(@Int) -> @Int.0 } }
```

```
$ vera check  main.vera   # OK
$ vera verify main.vera   # OK — "Verification: 6 verified (Tier 1)"
$ vera run    main.vera   # exit 1: Postcondition violation in mk(@Int -> @Box) ensures(@Box.result == <expr>) failed
```

The postcondition is genuinely true (the result IS `MkBox(input)`), Z3 proves it structurally, but codegen's runtime check does a pointer compare → spurious trap.

## Root-cause pointer

The `ResultRef` operand of `==` in an `ensures` clause is unhandled by `_infer_vera_type`, so a composite/ADT postcondition equality lowers to raw `i32.eq`. The fix: lower `==`/`!=` on composite types **in postcondition (and general) position** to the same **structural equality** codegen already uses for `eq(a, b)` on ADTs (it recurses fields and works at runtime). Cover Option/Result/Tuple/user-ADT operands, and both operand orders (`@T.result == ctor` and `ctor == @T.result`).

## Verification method

This is exactly the `ensures(P)` + `vera run` differential: the fix must make `vera run` on the repro succeed (print 5) while `vera verify` still proves it, AND a genuinely-false composite postcondition must still be rejected by both verify (E500) and a runtime trap. Include that negative control.

## Scope

Pre-existing; surfaced by the check-green→codegen-fail hunt (the only severe / soundness-adjacent finding of the 23). **Highest priority.** Fix for v0.1.0 on `release/v0.1.0`.

aallan:
## Summary

Fixes #912 (closes on the final `release/v0.1.0` → `main` merge).

**SEVERE soundness bug:** a composite (user-ADT / `Option` / nested-ADT) equality in an `ensures` postcondition that `vera verify` certifies at Tier 1 **traps at runtime on its own proved postcondition**.

### Repro

```vera
private data Box { MkBox(Int) }
private fn mk(@Int -> @Box) requires(true) ensures(@Box.result == MkBox(@Int.0)) effects(pure) { MkBox(@Int.0) }
public fn main(@Unit -> @Int) requires(true) ensures(true) effects(pure) { match mk(5) { MkBox(@Int) -> @Int.0 } }
```

| Command | Before | After |
|---|---|---|
| `vera check` | OK | OK |
| `vera verify` | OK (Tier 1) | OK (Tier 1) |
| `vera run` | **exit 1 — `Postcondition violation in mk`** | **prints `5`, exit 0** |

The postcondition is TRUE (the result IS `MkBox(input)`) and Z3 proves it structurally, but codegen lowered the composite `==` in the *runtime* postcondition check to a raw `i32.eq` (a pointer compare). The freshly-built return value and the freshly-built contract operand are structurally equal at different heap addresses, so `i32.eq` returned false → a spurious trap on a proved contract.

## Root cause

The `@Box.result` (`ResultRef`) operand of `==` resolved to a `None` Vera type in `_infer_vera_type` (`vera/wasm/inference.py`) — it was in that walker's "cannot occur" set — so the structural-`==` dispatch in `vera/wasm/operators.py` (guarded on the operand resolving to a known ADT name) was skipped, and the code fell through to the scalar `i32.eq` default. `_infer_expr_wasm_type` already handled the `ResultRef` **width** (#891); only the Vera-type dispatch was missing.

## Fix

`_infer_vera_type` now resolves a `ResultRef` to its declared `@Type` (sharing the `SlotRef` name logic via a new `_ref_vera_type_name` helper), so a composite `@T.result == ctor` postcondition routes through the SAME structural-equality codegen that `eq(a, b)` and `ctor == ctor` already use — which recurses fields and enforces the contract correctly at runtime.

Two coupled gaps surfaced and were closed in lockstep so the **#732 verifier↔codegen differential stays green**:

- The structural-`==` dispatch now recognises a monomorphized generic-ADT clone that lost its type argument (the #772 residue: `@T.result` → bare `@Box.result` for a `Box<T>` clone) and falls back to the pre-#912 scalar lowering rather than raising a spurious `E613`. Genuinely non-`Eq` operands (`Map`/`Array`-field ADTs, `Tuple`) still raise the correct `E613`.
- An imported generic ADT's type-parameter metadata is now propagated alongside its layout in `vera/codegen/modules.py`, without which the cross-module `Box<T>` clone could not be recognised.

## Soundness (negative control)

A genuinely-FALSE composite postcondition (`ensures(@Box.result == MkBox(0))` while the body returns `MkBox(@Int.0)`) is **still rejected by `vera verify` at Tier 1 (E500 counterexample) AND traps at runtime** — verified both ways. The fix makes true composite postconditions pass, never false ones. The verifier's own equality modelling is untouched.

## Tests

- New run-level conformance program `ch06_composite_postcondition_eq` (suite now 131).
- `tests/test_composite_postcondition_eq_912.py`: user-ADT / `Option` / nested-ADT true-postcondition run tests in both operand orders, the E500 + runtime-trap negative controls, and a Tier-1 verify pin.
- **Mutation-validated**: neutering the `ResultRef` arm flips exactly the left-`@T.result` composite-run tests RED with `Postcondition violation`, while the negative controls stay rejected.
- Full gate green: `pytest -m "not stress"` (6479 passed), `mypy vera/`, `ruff check vera/` + `--select S`, conformance (131), `#732` differential (`tests/test_monomorphize_differential.py`), the structural-`Eq` `E613` suite, doc-counts, site assets, diagnostic fields.


aallan:
## Round 2 — fixed the codegen-crash regression on `fn(@Box<T> -> @Box<T>)`

Pushed `6586945`. The round-1 fix introduced a new uncaught `AdtEqNotDerivableError` on a valid program (a function generic over the parameterized ADT itself with a slot-vs-slot postcondition). Confirmed a regression vs base (base ran it to `7`).

**Approach: option 1 (free-type-var scalar-fallback), not concrete-resolve.** The monomorphizer does not specialize `id2` to a concrete clone here — its `@Box<T>` operands carry an *unresolved type variable* `T`, so there is no concrete `Box<Int>` to resolve to at the codegen site. The lost-type-arg guard now treats a name whose type argument is a free type variable (`Box<T>`) as non-dispatchable and falls back to the pre-#912 scalar lowering — distinct from a concrete non-Eq argument (`Box<Array<Int>>`), which still routes to the gate and raises the correct `E613`.

**Tier-1-vs-Tier-3 analysis (no #912 resurrection).** `@Box<T>.result == @Box<T>.0` for a free `T` defers to **Tier 3**, it is NOT structurally proved at Tier 1 (pinned by `test_generic_param_adt_ensures_is_tier3_not_tier1`, asserting `tier3_runtime >= 1`). Because there is no Tier-1 structural proof, a scalar (pointer) runtime check cannot contradict one — so the fallback is sound for this shape. For the identity function the two operands are the same pointer, so `i32.eq` is also *correct*, matching base. If a future change made this prove at Tier 1, that soundness pin flips and forces a concrete-structural lowering.

**Backstop (reviewer's option 2, additively).** A postcondition-path `except AdtEqNotDerivableError` in `vera/codegen/functions.py` now catches a *genuinely* non-Eq contract composite (`Tuple`, or a concrete `Array`/`Map`-field argument) and emits the same clean `E613` the body/closure paths already emit, instead of the uncaught traceback the `ensures` path produced. The provable free-type-var case never reaches this backstop (routed to scalar before dispatch), so it does not cause a spurious E613 on a provable program.

**Evidence.** RED-first (both `==` and `!=` forms crashed on the prior HEAD, run after). All round-1 tests + the negative controls (verify E500 + runtime trap) stay green. #732 differential + structural-Eq E613 suite green. Full gate: `pytest -m 'not stress'` = 6405 passed / 53 skipped / 0 failed; mypy, ruff + `--select S`, conformance (131), doc-counts, site-assets, diagnostic-fields, examples, encoding all clean. Mutation-validated both round-2 pieces independently (neuter free-type-var routing → `Box<T>` run tests RED; remove backstop → clean-E613 tests raise uncaught).

aallan:
## Round 3 — corrected the soundness rationale + replaced the vacuous pin

Pushed `74fb4c0` (docs/tests only; no functional code change — the fix is unchanged from round 2). Thank you for catching this — accuracy matters most here.

**The rationale was factually wrong.** Enumerating obligations directly confirms your finding: the `id2`/`rebox` `ensures` is `status=verified` at **Tier 1** (the verifier substitutes `T:=Int` and proves the composite `==` structurally). The only Tier-3 obligation is an incidental `nat_to_int_coerce` in `main`, unrelated to the `==`. My "defers to Tier 3" claim was wrong.

**The TRUE rationale (verified from emitted WAT), now in the comment / docstring / CHANGELOG:** the scalar-lowered **base generic clone** `$id2` (`i32.eq`) is emitted but is **dead code** — never a call target, never exported (a bare generic fn can't escape higher-order; that's a parse error `E005`, so it can't reach a `call_indirect`/table). Every *reachable* call dispatches to a **monomorphized clone** `$id2$Int` whose `@Box<Int>.result == @Box<Int>.0` is lowered **structurally** (`call $eq_Box_LInt_R`; `Box<Int>` is concrete, not matched by the free-var guard), correctly discharging the Tier-1 proof. The free-var scalar fallback merely lets the dead base clone *compile* (as harmless dead code) instead of `E613`-erroring and failing the whole compile.

**Replaced the vacuous `..._is_tier3_not_tier1` pin** (it asserted `tier3_runtime >= 1`, passing on the incidental coercion; it stayed GREEN under the free-var mutation) with two real pins:
- `test_rebox_fresh_pointer_result_is_tier1_verified_and_runs` — a generic `fn(@Box<T> -> @Box<T>)` whose body returns a **freshly-constructed, structurally-equal, DIFFERENT-pointer** box; asserts the composite `==` is proved at **Tier 1** AND runs without trapping (the exact case that would trap under a reachable scalar lowering).
- `test_reachable_mono_clone_postcondition_is_structural` — asserts the reachable mono clone's postcondition WAT contains `call $eq_`, not `i32.eq`.

**Mutation evidence (two independent, load-bearing witnesses):**
- Neuter the **free-var routing** → the two `genslot` run tests flip RED (dead-base-clone `E613`); the `rebox` pins stay GREEN (that mutation only touches the dead base clone).
- Neuter the **`ResultRef` arm** → the `rebox` Tier-1+run pin AND the structural-WAT pin flip RED (the reachable mono clone becomes scalar and traps). *This is the mutation the reboxes guard* — the deeper invariant that the reachable path is structural.

Full gate green: `pytest -m 'not stress'` = 6406 passed / 53 skipped / 0 failed; mypy, ruff + `--select S`, conformance (131), doc-counts (6485), site-assets, diagnostic-fields, examples, changelog. No `KNOWN_ISSUES` change.

Issue #775, opened 2026-06-21
Monomorphized WAT name mangling is non-injective (_mangle_fn_name collides distinct instantiations)


aallan:
## Summary

`Monomorphizer._mangle_fn_name` (`vera/monomorphize.py`) builds WAT symbol names by **lossy** replacement — `ct.replace("<", "_").replace(">", "").replace(", ", "_")`, joined with `_` — so the mapping from instantiation to symbol is not injective. A generic instantiated at a parameterized built-in and at a user type whose name coincides with the mangled form collide to the same `$`-suffixed symbol:

- `g<Map<String, Int>>` → `g$Map_String_Int`
- `g<Map_String_Int>` (a user ADT literally named `Map_String_Int`) → `g$Map_String_Int`

Multi-parameter joins can also collide across the `_` boundary: `g<A_B, C>` vs `g<A, B_C>` both → `g$A_B_C`.

## Impact

**Codegen only.** The verifier keeps the original `decl.name` for its clones (#732) and never mangles, so this is *not* a verification-soundness issue — `vera verify` is unaffected. On the codegen side a collision either (a) trips a loud duplicate-symbol error, or (b) routes both instantiations to the first-registered clone (silently wrong). Which one occurs depends on whether Pass 1.5 clone emission has collision detection on the mangled name; needs confirming.

## Pre-existing

`_mangle_fn_name` was relocated **verbatim** from `vera/codegen/monomorphize.py` by #732; the non-injectivity predates it. Surfaced by the PR #767 review (CodeRabbit). It is out of scope for #732: a robust fix (escaped, length-prefixed component encoding) changes every monomorphized WAT symbol, churning the byte-identical-WAT baseline that PR is gated on.

## Fix

Make the encoding injective — escape `_` / `<` / `>` / `,` distinctly, or use length-prefixed components (`g$3Map6String3Int`-style), so distinct instantiations never share a symbol.

## Related

- #732 / #767 — where surfaced.
- #769 — sibling monomorphizer-precision gaps.

Issue #911, opened 2026-07-03
Show/Hash abilities unimplemented in codegen for composite types (ADT/Tuple/Option/Result/Array): check-green, function dropped at compile


aallan:
## Summary

The `Show` and `Hash` abilities are registered as **fully polymorphic** builtins (`vera/environment.py`: `show: (TypeVar A) -> String`, `hash: (TypeVar A) -> Int`), so the checker accepts `show(x)` / `hash(x)` for **any** type (spec §9.8 documents Show/Hash as universal). But WASM codegen implements them **only for primitives** — `vera/wasm/calls_handlers.py` `_translate_show` (lines 30–73) and `_translate_hash` (lines 75–119) dispatch via `_SHOW_DISPATCH` (Int/Nat/Bool/Byte/Float64 + special-cased String/Unit/Decimal) and `raise CodegenSkip(arg, "show()/hash() not supported for type <T>")` for every composite. The enclosing function is silently dropped from the WASM exports → `vera check` passes but `vera run` fails.

This is **check-green → codegen-fail** on ordinary code. It affects `show`/`hash` on: user ADTs (nullary-enum and field-carrying), `Tuple`, `Option`, `Result`, `Array`, and nested ADTs — **14 confirmed variants** (systematic hunt).

Note: structural `eq`/`==` on ADTs already works (codegen recurses fields), so the machinery to walk composite layouts exists — `show`/`hash` were simply never extended to composites.

## Canonical repros

```
private data Foo { MkFoo(Int) }
public fn main(@Unit -> @String) requires(true) ensures(true) effects(pure) { show(MkFoo(5)) }
```
```
private data Color { Red, Green, Blue }
public fn main(@Unit -> @Int) requires(true) ensures(true) effects(pure) { hash(Red) }
```
Both: `vera check` → OK; `vera run` → `Function 'main' body contains unsupported ... : show()/hash() not supported for type 'Foo'/'Color' — function skipped.` then `No exported functions to call.` (`show(42)`/`hash(7)` on primitives work — the control.)

## Fix direction

Implement `_translate_show` and `_translate_hash` for composite types by recursing over the value's layout, mirroring how structural `eq` already walks ADT fields:
[...]

PR #919, opened 2026-07-03
fix(codegen): effect handlers + effect-op results over composite type args (#914)

### LINKED ISSUE #914: Effect handlers (State/Exn) over composite/parameterized type arguments fail at codegen: check-green, invalid WAT / unknown func / ICE
## Summary

Effect handlers (`State`, `Exn`) and effect-operation results **over composite / parameterized type arguments** (`State<Tuple<…>>`, `State<Option<…>>`, `Exn<Tuple<…>>`, `Exn<Option<…>>`, an effect-op result of a composite type) fail at codegen while `vera check` passes. **6 confirmed variants** (systematic hunt), several distinct root causes under one theme:

1. **Unescaped type args in emitted WAT names.** A `State<Tuple<Int, Int>>` handler emits a host-import name `state_get_Tuple<Int, Int>` and an `Exn<Tuple<Int, Int>>` handler emits a tag `$exn_Tuple<Int, Int>` — the `<`, `>`, and space are illegal in WAT identifiers → invalid WAT. (Same escaping class as the injective-mangler work; the effect-name path was missed.)
2. **Missing helper generation.** A `State<Option<Int>>` handler references a `state_push_Option` helper that is never defined → `unknown func`.
3. **Dangling `@T` in an Exn payload binding.** An `Exn<Option<Int>>` handler binding `@Option<Int>` from the throw payload triggers an internal compiler error (dangling `@T.n`) at codegen.
4. **Effect-op result in a constructor argument.** An effect-op call result (`State.get()`) used directly as a `Tuple`/ADT constructor field is `CodegenSkip`-ped (the effect-op-result type isn't inferred at the constructor-arg site — same family as #905 for module calls).
5. **Inline match on an effect-op call** with a non-primitive op result — codegen cannot infer the match scrutinee WASM type.

Each is `vera check` OK → `vera run`/`vera compile` fails (invalid WAT, `unknown func`, ICE, or dropped function).

## Root-cause pointers

- Effect-handler codegen name construction (State/Exn host-import + tag names) must run type-argument names through the injective escaper (no raw `<`/`>`/space) — the same normalization used for monomorphized function names.
- `state_push`/`state_get` (and Exn tag/payload) helper generation must cover parameterized/composite state types, and the Exn payload binding must substitute `@T` to the concrete composite type (no dangling slot).
- Effect-op result WASM/Vera type inference must be resolved at constructor-argument and match-scrutinee positions (mirror the #905 constructor-field resolution and the #902/#905 inference fixes).

## Scope

Pre-existing; surfaced by the check-green→codegen-fail hunt. Fix for v0.1.0 on `release/v0.1.0`. May land as more than one PR (escaping vs helper-gen vs effect-op-result inference are separable); all reference this issue.

aallan:
Fixes #914 (closes on the final `release/v0.1.0` → `main` merge).

Effect handlers (`State`/`Exn`) and effect-op results over a **composite / parameterized type argument** passed `vera check` but failed at codegen. Six confirmed shapes, three root causes. Primitive `State<Int>` / `Exn<Int>` (the common case) are untouched.

## The six repros (before → after)

| # | Program (essence) | Before | After |
|---|---|---|---|
| A1 | `let @Tuple<Int, Int> = Tuple(get(()), 1); 0` under `<State<Int>>` | `could not infer constructor argument WASM type` — function skipped, `main` not exported | runs → `0` |
| A2 | `match get(()) { Box(@Int) -> @Int.0 }` inside `handle[State<Box>]` | `could not infer match scrutinee WASM type` — skipped | runs → `0` |
| B1 | `handle[State<Tuple<Int, Int>>]` | `WAT compilation failed: expected \`)\`` at `$vera.state_get_Tuple<Int, Int>` | runs → `0` |
| B2 | `handle[State<Option<Int>>]` get+put | `unknown func: failed to find name \`$vera.state_push_Option\`` | runs → `7` |
| B3 | fn declaring `<Exn<Tuple<Int, Int>>>`, `throw(Tuple(1, 2))` | `WAT compilation failed` at `(tag $exn_Tuple<Int, Int> …)` | compiles; uncaught throw traps at run (like primitive `Exn<Int>`) |
| C | `handle[Exn<Option<Int>>] { throw(@Option<Int>) -> option_unwrap_or(@Option<Int>.0, 0) } in { throw(Some(9)) }` | `[E699]` dangling `@T.n` | runs → `9` |

## Root causes and fixes

**A — effect-op RESULT type not inferred in ctor-arg / match-scrutinee position** (A1, A2). A bare `get(())` (a `State<T>` op, an `ast.FnCall`) used directly as a constructor field or match scrutinee left `_infer_expr_wasm_type` returning `None`, so the enclosing function was silently skipped. The op's result WAT type is now recorded in a new `_effect_op_result_wt` registry at every injection site (declared-effect path in `vera/codegen/functions.py`, handler-body path in `vera/wasm/calls_handlers.py`) and consulted by both the `FnCall` and `QualifiedCall` inference arms in `vera/wasm/inference.py`. (The old `QualifiedCall` arm's `_fn_ret_types` lookup on the dispatch target `$vera.state_get_T` was dead for *every* value-producing op.)

**B — handler WAT names not escaped for a composite type arg + an emit↔call desync** (B1, B2, B3). Every `state_*` / `exn_*` WAT identifier now routes its type-argument component through the injective `mangle_type_name` (#775) at all sites — `vera/codegen/assembly.py` import/tag decls, `vera/codegen/functions.py`, `vera/wasm/calls_handlers.py`, `vera/codegen/contracts.py`'s `old(State<T>)` read, and the host registration in `vera/runtime/state.py` (the `ExecuteResult.state` cell key stays the readable `State_<name>`). The handler body now uses the **full** canonical slot name (`_type_expr_to_slot_name`), so the `(import …)` decl and the body call can never drift (fixes the `state_push_Option` vs `Option<Int>` desync).

**C — Exn payload `@T` binding not substituted** (C). The `Exn` handler pushed its caught-payload binding under the base name (`Option`), so `@Option<Int>.0` in the handler body resolved to no local. It now binds under the full canonical slot name.

## Tests

- New run-level conformance programs `ch07_state_composite` and `ch07_exn_composite` (suite now 133).
- `TestEffectCompositeTypeArgs914` in `tests/test_codegen_effects.py` — all six shapes + two primitive-effect regression pins (`State<Int>`, `Exn<Int>`).
- A composite-`State` Node.js parity test in `tests/test_browser.py`.
- **Mutation-validated** independently: reverting the result-WT inference flips exactly A1/A2 RED; reverting the full-name selection reproduces the `unknown func $vera.state_push_Option` desync; reverting the payload binding reproduces the `E699` dangling slot — the three fixes are independently load-bearing, and the primitives stay green throughout.

Full non-stress suite green (6429 passed), `mypy`, `ruff` (+`--select S`), conformance (133), doc-counts, site-assets, diagnostic-fields, encoding, and the changelog gate all pass. A latent browser-only quirk (composite-Float `State` default seeding) is out of scope and flagged as a follow-up.


aallan:
## Round 2 — adversarial-review findings fixed + rebase

Rebased onto `release/v0.1.0` tip `15999a8` (#911/#912/#916 landed and touch the same `functions.py` / `inference.py` I edit). Conflicts were doc-count only; the code files auto-merged; full non-stress suite re-run against the combined tree: **6508 passed**, mypy + ruff (+`--select S`) clean, conformance **136**.

### Finding 1 — `old(State<Composite>)` crashed codegen (and the round-1 `contracts.py` mangle was dead for composites)

Root cause: `_extract_state_type_name` returned the BASE name (`Option`) for the `old(State<Option<Int>>)` snapshot lookup, missing the canonical-keyed `_state_types` entry (`Option<Int>`) — no snapshot local allocated, so the `old` read raised an uncaught `CodegenInvariantError` at run. Fixed by returning the canonical slot name. A downstream gap then surfaced: `_infer_expr_wasm_type` had no `OldExpr`/`NewExpr` case, so a composite `new(State) == old(State)` emitted an ill-typed `i64.eq` on two i32 pointers; it now resolves `State<T>`'s WAT type so the `==` lowers at i32. **`old(State<Option<Int>>)` now `vera run`s and `vera verify`s** (new conformance `ch07_state_old_composite` + `test_old_state_composite_*`).

### Finding 2 — nested composite type args collided (injectivity only half-true) + Finding 3 dedup

Six duplicated `_type_expr_to_slot_name` copies plus the `SlotRef`-lookup and closure-capture name builders descended only ONE level, so `State<Option<Tuple<Int, Int>>>` and `State<Option<Tuple<Bool, Bool>>>` both collapsed to the slot name `Option<Tuple>` → the same `state_*` import / `exn_*` tag (an injectivity hole; a type-confusion for Exn), and a checker↔codegen/verifier desync since the checker already recursed. Consolidated onto ONE shared recursive `vera.slots.type_expr_slot_name` / `slot_ref_name` that recurses fully into nested type args and matches the checker; `codegen/core.py`, `wasm/inference.py`, `smt.py`, `verifier.py`, `tester.py`, the `operators.py` / `smt.py` SlotRef builders, `wasm/closures.py`, and `_type_expr_name` all delegate to it. Two distinct nested-composite types now get distinct, correctly-matched names (WAT-grepped in the tests). The recursion surfaced a one-level capture-key desync in `closures.py` (a captured `@Array<Array<Int>>` dropped `main` with `unknown func $step_nested`), fixed in lockstep — `ch05_capture_array_index` pins it.

Two smt/tester coverage tests that pinned the pre-dedup fallbacks (`FnType`→`None`, non-`NamedType`-arg→`"?"`) are updated to the shared function's consistent values (`FnType`→`"Fn"`, refinement-arg→base name), aligning the tester/smt with the rest of the compiler.

Each fix independently **mutation-validated**: reverting the `old`-state canonical-name fix reproduces the crash; reverting the recursion collides the two nested-composite names AND reproduces the `unknown func $step_nested` closure desync. The original 6 gaps + primitive `State<Int>`/`Exn<Int>` pins stay green throughout.

Issue #900, opened 2026-07-03
Generic instantiated at Unit (forall<T> over @Unit) is check-green but fails/crashes at codegen


aallan:
**Found during the PR #899 (#878) review; pre-existing on both the release/v0.1.0 branch and main (not introduced by #878).**

## Symptom
A generic instantiated at the zero-size `Unit` type passes `vera check` but fails at codegen:

    private fn mku(@Int -> @Unit) requires(true) ensures(true) effects(pure) { () }
    private forall<T> fn idt(@T -> @T) requires(true) ensures(true) effects(pure) { @T.0 }
    public fn main(@Unit -> @Int) requires(true) ensures(true) effects(pure)
    { let @Unit = idt(mku(5)); 7 }

- `vera check` → OK.
- `vera run` → **fails, and the exact failure is hash-seed-dependent**: observed `E699` (function skipped), "No exported functions to call" (`main` dropped), and — worst — a `codegen invariant violation — … the type checker should have rejected the input before it reached this point. Please file a bug report` crash.

The crash message is itself telling: the codegen invariant expects the checker to have already rejected `forall<T>` at `T = Unit`.

## Nature
Instantiating a generic at `Unit` is a degenerate/niche case (identity over a zero-size value). The right fix is likely a **checker-side clean rejection** (a proper diagnostic that generics may not be instantiated at `Unit`, if that's the intended design) rather than a codegen change — OR codegen support for the `Unit` monomorphization. Either way, a check-green program must not reach a "file a bug report" codegen crash.

Pre-existing; surfaced by an edge-shape probe during #878's clone-naming work. Related: #878 (the user-fn-return-into-generic-arg clone naming this sits adjacent to).

Issue #922, opened 2026-07-04
Non-Eq composite == / hash / show in a contract (requires/refinement/ensures) escapes as an uncaught Python traceback at compile


aallan:
## Summary

A contract predicate (`requires`/`ensures`/refinement-type guard) that uses a non-Eq-derivable composite `==`/`!=`, or `hash`/`show` on an unsupported type, escapes as an **uncaught Python traceback** at compile (`vera compile`/`run`/`test`) instead of a clean diagnostic. The graceful degradation exists at SOME contract sites but not all — three sibling `translate_expr` call sites lack the `except AdtEqNotDerivableError`/`except CodegenSkip` the body/postcondition paths have.

## Repros (all `vera check` OK → `vera compile`/`run`/`test` uncaught traceback)

**G3 — non-Eq composite `==` in `requires()`:**
```vera
private fn f(@Int -> @Int) requires(Tuple(@Int.0, @Int.0) == Tuple(@Int.0, @Int.0)) ensures(true) effects(pure) { @Int.0 }
public fn main(@Unit -> @Int) requires(true) ensures(true) effects(pure) { f(7) }
```
→ `vera.skip.AdtEqNotDerivableError: ADT equality on non-Eq-derivable type 'Tuple<Int, Int>'` (uncaught). Site: `vera/codegen/functions.py:351` calls `_compile_preconditions` → `contracts.py:522` `translate_expr` with NO surrounding try/except (unlike body `functions.py:425`, closure `:522`, postcondition `:582` — the #912 fix).

**G4 — non-Eq composite `==` in a refinement-type parameter predicate:** the refinement param-guard emit path has the same missing catch → same uncaught `AdtEqNotDerivableError`.

**G5 — `hash()`/`show()` on an unsupported type in `ensures()`:** the postcondition catch (`functions.py:580-584`) only catches `AdtEqNotDerivableError`, NOT `CodegenSkip` — so `ensures(hash(recursiveADT) == 0)` escapes as an uncaught `CodegenSkip` traceback (the `contracts.py:197-198` catch is also narrow).

## Root cause + fix

[...]

Issue #841, opened 2026-07-01
Host-threaded concurrent <Async> for commutative effect rows


aallan:
## Summary

Make `async()`/`await()` actually concurrent for host-side effects — the concurrency half that #59 deliberately deferred when it closed at v0.0.82 (language surface only: `Async` marker effect, WASM-transparent `Future<T>`, `async`/`await` lowered as identity → eager, spec §9.5.4).

This is **host-threaded** concurrency: the guest stays single-threaded; only whitelisted host operations overlap. It requires nothing from WASI 0.3 (#406) and ships before it.

## Design (from the server-effects sprint plan)

- **Lowering** (`vera/wasm/calls_markup.py`): `async(e)` where `e` is a *direct call to a whitelisted host op* fuses into a single host import (e.g. `vera.async_http_get(ptr, len) -> i32`) that submits the work to a host `ThreadPoolExecutor` and returns a `Future` handle immediately. Every other `async(e)` shape stays eager (documented). `await` on a pending handle lowers to `vera.async_await(handle)`, which blocks on the guest thread and builds the result ADT there.
- **`Future<T>` representation**: a host handle using the #578 bit-31 tagging pattern (as Decimal does), so the conservative GC never mistakes it for a heap pointer; unawaited futures are reclaimed via the existing wrap-table `host_decref_handle` path with a new kind.
- **GC safety**: workers run pure host work (urllib) and never touch guest memory or the shadow stack; all allocation happens at `await` time on the (single) guest thread under `_ShadowGuard` — the standard, `test_codegen_gc_rooting.py`-covered pattern.
- **Checker rule (soundness-critical)**: `async(e)` is concurrency-eligible only when `e`'s inferred effect row is within a commutative whitelist (`{Http}` in v1). Effect rows do not currently restrict `State<T>` or `IO` under `Async`, so without this rule concurrent evaluation would make stdout/state re-ordering observable and silently break the verifier's sequential-semantics assumptions. The rule lands in the mutation-tested soundness core with RED-first, mutation-surviving tests.
- **Spec**: §9.5.4 rewritten — an implementation MAY evaluate `async(e)` concurrently when the effect row is commutative; value semantics unchanged; effects outside the whitelist retain program order. Verifier claims remain literally true.
- **Browser**: `runtime.mjs` keeps eager evaluation, which the new spec wording makes conformant (a sync JS import cannot block on fetch). Parity tests assert values only.
[...]

PR #940, opened 2026-07-04
fix(checker,codegen): reject generic @T read in a contract clause at Unit (E206), not a codegen crash (#939)

### LINKED ISSUE #939: Generic @T read in a requires/ensures clause crashes codegen at Unit instantiation (check + verify pass)
A `forall<T>` function that reads a `@T.n` slot **inside a `requires`/`ensures` contract clause** (not the body), instantiated at the zero-size `Unit` type, passes `vera check` and `vera verify` but **crashes at codegen with a raw `CodegenInvariantError` traceback** — violating the DESIGN.md invariant that a `check`-green program must never surface a raw Python traceback.

This is the same bug class as #900 (generic-over-`Unit` dangling slot), reached through the one door #900 did not close: #900's E206 checker-gate walks only `decl.body`, on the premise that a `@T` in a contract clause "never lowers to body WASM". That premise is false — `_compile_preconditions` (and the postcondition path) lower contract expressions to runtime WASM checks, where a `@T.n` read at `Unit` resolves to no local and dangles.

## Reproduce

```vera
private forall<T> fn ignore_pair(@T, @T, @Int -> @Int)
  requires(@T.1 == @T.0)
  ensures(true)
  effects(pure)
{
  @Int.0
}

{
  let @Int = ignore_pair((), (), 5);
  ()
}
```

- `vera check`  → **OK** (exit 0)
- `vera verify` → **OK** (exit 0)
- `vera run` / `vera compile` → **raw `CodegenInvariantError` traceback** (exit 1):

```
File "vera/wasm/operators.py", line 38, in _translate_slot_ref
    raise CodegenInvariantError("slot reference resolved to no local (dangling @T.n)", ref)
vera.skip.CodegenInvariantError: slot reference resolved to no local (dangling @T.n)
```

A control instantiating the same function at `Int` (`ignore_pair(3, 3, 5)`) checks, verifies, and runs cleanly — the crash is `Unit`-specific.

## Root cause

Two adjacent gaps, both in the "same class as #900" family:

1. **Checker (E206 scope):** `_forall_vars_read_in_body` (`vera/registration.py`) walks only `decl.body`, so a `@T.n` read confined to a `requires`/`ensures` clause is not counted, and the E206 gate (`vera/checker/calls.py`) never fires.
2. **Codegen (net asymmetry):** the precondition-compile `try/except` (`vera/codegen/functions.py`) catches only `(AdtEqNotDerivableError, CodegenSkip)` — the two exception types #922 added — but **not** `CodegenInvariantError`, which the sibling body and postcondition paths both catch and degrade to a clean `[E699]`. So the dangling-slot invariant escapes uncaught.

## Fix direction

Mirror #900's body path on both layers: (1) extend the E206 discriminator to also walk `requires`/`ensures` for `@T.n` reads (clean check-time diagnostic), and (2) add `CodegenInvariantError` to the precondition-compile catch so all three contract-lowering paths (precond / postcond / body) degrade uniformly to `[E699]` — completing the symmetry #922 began.

Found during the final pre-`v0.1.0` review sweep.

aallan:
Fixes #939 (closes on the final `release/v0.1.0` → `main` merge via #938, whose closing list is updated to include it).

Found during the final pre-`v0.1.0` review sweep: an adversarial probe of the #900 E206 fix surfaced a check-green program that still crashes codegen.

## The bug

A `forall<T>` function that reads a `@T.n` slot **inside a `requires` / `ensures` contract clause** (not the body), instantiated at the zero-size `Unit` type, passes `vera check` **and** `vera verify` but crashes `vera run` / `vera compile` with a raw `CodegenInvariantError` traceback — violating the DESIGN.md invariant that a `check`-green program never surfaces one.

```vera
private forall<T> fn ignore_pair(@T, @T, @Int -> @Int)
  requires(@T.1 == @T.0) ensures(true) effects(pure)
{ @Int.0 }

public fn main(@Unit -> @Unit) requires(true) ensures(true) effects(pure)
{ let @Int = ignore_pair((), (), 5); () }
```

`vera check` → OK · `vera verify` → OK · `vera run` → raw `slot reference resolved to no local (dangling @T.n)` traceback. A control at `Int` runs clean, so it's `Unit`-specific.

## Root cause

Same class as #900 (generic-over-`Unit` dangling slot), reached through the one door #900 left open — two adjacent gaps:

1. **Checker:** #900's E206 discriminator (`_forall_vars_read`) walked only `decl.body`, on the premise that a `@T` in a contract clause "never lowers to WASM". False — `requires` / `ensures` lower to the *runtime* pre/post-condition checks in `_compile_preconditions` / `_compile_postconditions`, so a `@T.n` read there dangles at Unit identically.
2. **Codegen:** the precondition-compile `try/except` caught only `(AdtEqNotDerivableError, CodegenSkip)` — the two #922 added — but **not** `CodegenInvariantError`, which the sibling body / closure / postcondition paths all catch and degrade to a clean `[E699]`.

## Fix (mirrors #900's body path on both layers)

- **`vera/registration.py`** — `_forall_vars_read` now also walks `requires` / `ensures` clauses (but **not** the verifier-only `decreases` / `invariant`, which never emit a `local.get` — counting those would over-reject a program that compiles fine). E206 now fires at check with a clean diagnostic. The discriminator stays aligned 1:1 with codegen: `_is_trivial_contract` skips **only** literal `true`, so every non-trivial contract — every one containing a real `@T.n` — is compiled, exactly the set the checker now rejects.
- **`vera/codegen/functions.py`** — the precondition-compile path now catches `CodegenInvariantError` → clean `[E699]`, completing the sibling-path symmetry #922 began. All three contract-lowering paths (precondition / postcondition / body) now catch the identical exception set.

## Validation

- Two checker tests (`requires` and `ensures` `@T`-at-Unit → E206) and one codegen test (forced precond invariant → `[E699]`, not a raw traceback), **both mutation-validated RED-first**.
- Over-rejection ruled out: the `@T`-unread accept-cases (`firstInt`, `ignore`, boxed `Option<Unit>`, built-in `async` over Unit) all still pass; the `Int` control still compiles and runs.
- Full non-stress suite 6,678 passed · conformance 143 · examples 37 · mypy / ruff (+`--select S`) / doc-counts / site-assets / diagnostic-fields all clean.

## CI note

PRs into `release/v0.1.0` don't run the CI matrix (CI triggers on PRs into `main`); this change is validated by the full local gate above and rides the matrix on the final `release/v0.1.0` → `main` PR (#938).


aallan:
Extended in d4c8496 after the completeness skeptic in the pre-`v0.1.0` adversarial review found this fix **incomplete**.

**Residual crash it caught:** a generic reading `@T` at `T = Future<Unit>` (pinned via `async(())`) — `check`-green **and** `verify`-green — still crashed `vera run` with a raw `CodegenInvariantError`. `Future<Unit>` is a *second* zero-size type: codegen makes `Future<T>` transparent to `T` (#841), so it erases to no WASM local exactly like bare `Unit`, but the E206 discriminator keyed on `base_type(b) == UNIT` and missed it. Separately, the *postcondition* codegen path lacked the `CodegenInvariantError` → `[E699]` net (the first commit added it only to the precondition; body/closure already had it), so a `@T` read in an `ensures` clause escaped uncaught.

**Closed by:** a shared `erases_to_unit` predicate (`vera/types.py`) that mirrors codegen's Future-transparency — so `Future<Int>` (i32) and a boxed `Option<Unit>` stay accepted (no over-rejection) — plus the postcondition net. All four contract-lowering paths (precondition / postcondition / body / closure) now degrade to a clean `[E699]`. +4 tests, each mutation-validated; the full non-stress suite, conformance (143), and examples (37) stay green.

Issue #730, opened 2026-06-12
Verifier: calls in statement position are never precondition-checked


aallan:
A violating call whose value is discarded — a bare statement like consume(need_pos(@Nat.0)); — produces no E501: the body translation skips ExprStmt nodes entirely, and the @Nat-subtraction walker only translates expressions at let RHSes, branch conditions, and Nat-subtraction operands. So the precondition of a call in statement position is checked only when the call happens to sit inside one of those shapes.

Repro (identical on main and the issue-727 fix branch — this is a pre-existing coverage gap, surfaced by the probe matrix used to validate the dedup fix in the PR closing 727/728):

    private fn need_pos(@Nat -> @Nat) requires(@Nat.0 >= 1) ensures(true) effects(pure) { @Nat.0 }
    private fn consume(@Nat -> @Nat) requires(true) ensures(true) effects(pure) { @Nat.0 }
    public fn caller(@Nat -> @Nat) requires(true) ensures(true) effects(pure) {
      consume(need_pos(@Nat.0));
      @Nat.0
    }

vera verify reports no E501 (expected: one, at the need_pos call site). The same call as a block value or let RHS is checked correctly.

Fix direction: translate (or at least walk-and-check calls in) ExprStmt subtrees during the body pass — with the identity dedup recording shipped for issue 727, re-translation is safe and cannot reintroduce duplicates.

Tracked in the KNOWN_ISSUES.md Limitations table.

Issue #727, opened 2026-06-11
Verifier emits duplicate E501 diagnostics for a violating call site


aallan:
A call site that violates the callee's precondition produces TWO byte-identical E501 diagnostics (same line, column, description, counterexample) in vera verify --json, and consequently a doubled entry in the LSP hover/Problems panel.

Repro (v0.0.168): take examples/inference.vera and change the call in main to classify_sentiment("") — the empty string violates classify_sentiment's requires(string_length(@String.0) > 0). Then:

    vera verify --json examples/inference.vera | jq '[.diagnostics[] | {code: .error_code, line: .location.line, column: .location.column}]'

yields the E501 entry twice at line 25 col 33. Observed identically in the VS Code extension's hover (two stacked copies of the same diagnostic).

Surfaced while screenshotting the new LSP integration; the doubled hover is cosmetically loud in an editor even though the CLI exit code and verification summary are unaffected. Likely a double-recording of the call-site precondition violation somewhere between drain_call_violations and the diagnostic stream; the fix should dedup at the source rather than in the LSP layer so --json consumers benefit too.

Tracked in the KNOWN_ISSUES.md Bugs table.

Issue #764, opened 2026-06-20
Call preconditions after a `let`-destructure are not statically checked (_translate_block truncates at LetDestruct)


aallan:
## Summary

A call-site precondition (`E501`) is **not statically checked** when the call appears at or after a `let`-destructure statement in a function body (or in the block's final expression when the block contains any destructure). The program verifies `tier1`-clean even though the precondition was never discharged.

This is a verification **completeness / classification** gap, not a silent wrong answer: the callee's own runtime `requires(...)` contract check still traps if the precondition is violated, so it should be reported `tier3_runtime`, not `tier1`. Pre-existing and refinement-independent — surfaced while reviewing #746 (PR #763).

## Root cause

`SmtContext._translate_block` (`vera/smt.py`) bails at the first non-`LetStmt`/`ExprStmt` statement:

```python
else:
    # LetDestruct or unknown statement type
    return None  # pragma: no cover
```

A `let Ctor<...> = ...;` destructure hits this branch, so the block translation returns `None`, truncating SMT translation of everything from the destructure onward — including the block's final expression. Call-site preconditions are collected during that main SMT body translation (`drain_call_violations`), so any call at/after the destructure is never seen. (The `# pragma: no cover` shows this path was assumed unreachable; a destructure followed by a call reaches it.)

## Reproduction

```vera
public fn needs_pos(@Int -> @Int)
  requires(@Int.0 > 0) ensures(true) effects(pure)
{ @Int.0 }

public fn mk(@Int -> @Tuple<Int, Int>)
  requires(true) ensures(true) effects(pure)
{ Tuple(@Int.0, 3) }

public fn f(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  let Tuple<@Int, @Int> = mk(@Int.0);
  needs_pos(@Int.1)            -- @Int.1 is unconstrained; requires(> 0) NOT checked
}
```

`vera verify` reports this clean at `tier1` (no `E501`). The same body with the call **before** the destructure correctly fires `E501`, and the no-destructure baseline correctly fires `E501` — confirming the destructure is what truncates the check.

## Scope (verified)

- **Affected:** call-site preconditions (`E501`) — and any check done via the main SMT body translation, e.g. a postcondition on the result — for statements at/after a `let`-destructure, or in the block's final expression when the block contains a destructure.
[...]

PR #953, opened 2026-07-06
fix(smt): check preconditions of calls nested in effect-op arguments (#776)

### LINKED ISSUE #776: Precondition of a call nested inside an effect-operation argument is not statically checked
## Summary

A precondition-bearing call passed as an **argument to an effect operation** (a `QualifiedCall` such as `IO.print(...)`) is never precondition-checked — no E501 fires, though `vera verify` reports clean. `translate_expr` has no dispatch branch for `QualifiedCall` and falls through to `return None` (`vera/smt.py`), so it never recurses into the effect op's arguments; the nested call is therefore never translated and its `requires(...)` is never checked statically.

## Repro

```vera
private fn fmt_pos(@Nat -> @String)
  requires(@Nat.0 >= 1)
  ensures(true)
  effects(pure)
{ "x" }

public fn caller(@Nat -> @Nat)
  requires(true)
  ensures(true)
  effects(<IO>)
{ IO.print(fmt_pos(@Nat.0)); @Nat.0 }
```

`vera verify` → `OK:` (clean), but `fmt_pos`'s `requires(@Nat.0 >= 1)` is violable (the caller only guarantees `true`). The same call *not* wrapped in an effect op — `need_pos(@Nat.0);` — correctly fires E501 after #730.

## Root cause

`translate_expr` dispatches `FnCall` / `ModuleCall` to the call-precondition path (`_translate_call_with_info`, which checks each `Requires`), but a `QualifiedCall` (effect operation) is intentionally not translated to a Z3 term — effects in *contract predicates* violate purity — so it returns `None` without recursing into its arguments. A call-precondition-bearing argument is collateral. (The `@Nat` / refinement narrowing walkers *do* descend into `QualifiedCall` args, so narrowing obligations at effect-op-arg sites are unaffected — only the E501 call precondition is missed.) The callee's runtime `requires(...)` still guards it, so this is a verification-completeness gap, not a silent wrong answer.

## Scope

Distinct from #730 (direct statement-position calls, now fixed) and #764 (a call at/after a `let`-destructure). This is the effect-op-argument facet of the same "calls whose preconditions escape static checking" family.

## Fix direction

Recurse into a `QualifiedCall`'s argument expressions for their call-precondition side effect (without translating the effect op itself to a Z3 term) — mirroring how the #730 fix translates `ExprStmt.expr` for its side effect, and how the narrowing walkers already descend into effect-op args.

## Related

- #730 (statement-position calls), #764 (`let`-destructure truncation), #754 / #757 (effect-op-arg runtime guards for `@Nat`).

### LINKED ISSUE #957: Verifier reports a false Tier-1 proof: a callee's ensures is assumed without the enclosing branch's path condition
## Summary

`vera verify` reports a **false Tier-1 proof**. It prints `OK` and counts a postcondition as `verified` for a program whose `ensures` is false and traps at run time.

The verifier assumes a callee's `ensures` onto the solver's **base** assertion stack, without the enclosing branch's path condition. `check_valid` folds `_path_conditions` around the *goal* only, so a fact learned under an `if` outlives the branch and becomes an unconditional fact about the **caller's own slots** (`_build_callee_env` binds the callee's parameters to the caller's terms).

## Reproduction (no effect operation anywhere)

```vera
private fn dec5(@Nat -> @Nat)
  requires(@Nat.0 >= 5)
  ensures(@Nat.result == @Nat.0 - 5)
  effects(pure)
{
  @Nat.0 - 5
}

public fn main(@Nat -> @Nat)
  requires(true)
  ensures(@Nat.0 >= 5)          -- FALSE: main accepts 0
  effects(pure)
{
  if @Nat.0 >= 5 then { let @Nat = dec5(@Nat.0); () } else { () };
  @Nat.0
}
```

```
$ vera verify vac_pure.vera
OK: vac_pure.vera
Verification: 6 verified (Tier 1)

$ vera run vac_pure.vera --fn main -- 0
Error: Postcondition violation in main(@Nat -> @Nat)
  ensures(@Nat.0 >= 5) failed
```

`vera verify` produces **more successes and fewer diagnostics** than a correct verifier — the signature of failing open.

## Why it is circular

`dec5`'s `ensures(@Nat.result == @Nat.0 - 5)` injects `ret == @Nat.0 - 5`. Combined with `@Nat`'s implicit `ret >= 0` (from `declare_nat`), Z3 derives `@Nat.0 >= 5` **globally** — the very precondition the branch guard existed to establish. `main`'s false `ensures(@Nat.0 >= 5)` then discharges.

## Vacuity: it also suppresses diagnostics

Two calls in mutually-exclusive arms inject contradictory facts about the same caller slot. The base solver becomes **UNSAT**, `check_valid`'s refutation query returns `unsat` for *every* goal, and every obligation discharges vacuously — including `E501` call-precondition diagnostics that should fire. A false Tier-1 does not merely add a wrong "verified"; it silently deletes real diagnostics.

## Root cause

`vera/smt.py`, `_translate_call_with_info`:

- `self.solver.add(z3_post)` — the callee's `ensures`
- `self.solver.add(z3_pred)` — a refined return's predicate

Neither consults `self._path_conditions`. `grep -n "_path_conditions"` returns no hit anywhere in the call-translation region.

**`_translate_match` already does this correctly** — `self.solver.add(z3.Implies(cond, f))`. The call translator is the outlier.

## Impact

`vera verify` is wrong. The compiled program still enforces the contract: `cmd_compile` never calls the verifier and `vera/codegen/contracts.py` has no notion of tiers, so the runtime `contract_fail` guard is emitted regardless of the tier. The failure therefore surfaces as a runtime trap rather than a compile-time diagnostic.

**This mitigation is accidental, not designed.** `spec/11-compilation.md` §11.8 states "Contracts that the verifier proved (Tier 1) are omitted from the compiled output". That optimisation is not implemented (tracked separately). If it is ever implemented before this bug is fixed, every false Tier-1 becomes a **silent wrong answer**.

## Fix

Scope each injected fact to the path that earned it:

```python
def _guard_fact(self, fact: z3.ExprRef) -> z3.ExprRef:
    if self._path_conditions:
        return z3.Implies(z3.And(*self._path_conditions), fact)
    return fact
```

applied at both `solver.add` sites. Same soundness as discarding the facts, strictly more precision, and it matches the `_translate_match` idiom already in the file.

## Validation

- the repro above, and the effect-op variant, both flip `ok=true` -> `ok=false` + `E501`/`E500`
- a legitimately guarded call (`if @Nat.0 >= 1 then { IO.print(need_pos(@Nat.0)) }`) is **not** over-rejected
- 180 conformance + example programs: `tier1 1464 -> 1464`, `tier3 594 -> 594`, zero diagnostic changes
- full suite green

Found by the adversarial review of [#953](https://github.com/aallan/vera/pull/953), which surfaced it because that PR routed effect-op arguments into the same unguarded mechanism and turned correctly-rejected programs into falsely-accepted ones.

chethanuk:
## Summary

A call's precondition (`requires(...)`) was **not statically checked** when the
call was nested inside an **effect-operation argument** — e.g. `IO.print(need_pos(@Nat.0))`.
`vera verify` reported clean and no `E501` fired, even though the nested callee's
`requires(@Nat.0 >= 1)` is violable when the caller only guarantees `true`.

**Root cause.** `translate_expr` (`vera/smt.py`) dispatched `FnCall` / `ModuleCall`
to the call-precondition path but had **no `QualifiedCall` branch**. An effect op
is deliberately *not* Z3-translated (an effect inside a contract predicate would
violate purity), so it fell straight through to `return None` — **without recursing
into its arguments**. A precondition-bearing call passed as an argument was therefore
never visited, so its obligation was never emitted. The callee's *runtime* `requires(...)`
still guards it, so this was a verification-completeness gap, not a wrong answer.

**Fix.** Add a `QualifiedCall` branch that walks each argument via `translate_expr`
purely for the `E501` precondition side effect, drops the value, and returns `None`
so the effect op itself is never Z3-translated. This mirrors the `#730`
statement-position (`ExprStmt`) handling; the `#727` span-keyed dedup keeps the
re-translation duplicate-free. The `@Nat`/refinement narrowing walkers already
descended into effect-op args, so only the `E501` call precondition was missed.

Worked example — `IO.print(need_pos(@Nat.0))` in a `requires(true)` caller:

```
translate_expr(QualifiedCall IO.print(...))
  |- (effect op: not translated to a Z3 term)
  |  for arg in args:  -> translate_expr(FnCall need_pos(@Nat.0))
  |                        |- emits call_pre obligation -> E501 (requires @Nat.0 >= 1 unmet)
  |- return None        <- effect op stays untranslated
```

## Related Issues

Closes #776

## Type of Change

- [ ] Specification change
- [x] Compiler implementation
- [x] Bug fix
- [x] Tests
- [x] Documentation

## Checklist

- [x] All tests pass locally

### How verified

- `tests/test_obligations.py` — new regression `test_precondition_checked_inside_effect_op_argument`
  asserts exactly one `E501` and one `call_pre` obligation for `IO.print(need_pos(@Nat.0))`.
  It is load-bearing: reverting the `QualifiedCall` branch drops the count to `0` and the test fails.
- `python -m pytest tests/test_obligations.py -q` -> **371 passed**.
- `mypy vera/smt.py` -> clean.
- `python scripts/check_limitations_sync.py` -> consistent (the now-fixed `#776` row was removed
  from the `KNOWN_ISSUES.md` limitations table).
- `CHANGELOG.md` gains an `[Unreleased] -> Fixed` bullet.

---

**Maintainer note (adversarial review).** This PR now also fixes the root-cause verifier bug it surfaced. `_translate_call_with_info` assumed each callee `ensures` — and each refined return's predicate — onto the solver's *base* assertion stack, unguarded by `_path_conditions`, so a fact learned inside an `if` escaped the branch and became an unconditional fact about the caller's own slots. `_translate_match` already guarded its injected facts this way; the call translator was the only site that did not.

Routing effect-op arguments into that mechanism turned programs `main` correctly rejected into ones it silently accepted. The same leak is reachable on `main` today with no effect operation anywhere, and its vacuity (contradictory facts from mutually-exclusive arms → UNSAT base solver) silently deleted real `E501` diagnostics, including the ones this PR adds.

Each injected fact is now wrapped in `z3.Implies(z3.And(*self._path_conditions), fact)`. Behaviour-neutral across all 143 conformance programs and 37 examples.

Closes #957

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved verification so branch-specific facts no longer leak into unrelated code paths.
  * Fixed checks so nested function calls are still validated inside operation arguments.
  * Strengthened handling of contract-related expressions to avoid incorrect proof results.

* **Tests**
  * Added coverage for nested-call validation, multi-argument operation handling, and branch-safe verification behaviour.

* **Documentation**
  * Updated project statistics and release metrics across the docs.
  * Refreshed the known issues list to reflect the latest fixed items.


aallan:
Thanks for this. It's the most valuable of your three PRs, and not only for the reason you filed it.

Alongside CodeRabbit I ran this through an adversarial review pointed specifically at verifier soundness — parallel finder lenses, each finding then attacked by independent skeptics prompted to refute by default, with a judge that re-verifies the survivors from scratch. For a change to `vera/smt.py` that seemed proportionate: a verifier that fails *closed* is an annoyance, and a verifier that fails *open* is the one bug this language cannot afford. It found something neither of us was looking for.

## Your fix is real, and `main` was worse than #776 said

#776 describes a missed diagnostic. It's stronger than that. On `main`:

```
$ vera verify c776.vera        # IO.print(need_pos(@Nat.0)), requires(@Nat.0 >= 1)
OK: c776.vera
```

`vera verify` reports **`ok=true`** for a program whose precondition can be violated. Your branch reports `E501`. That's not incompleteness — that's the verifier saying a program is fine when it isn't.

The `QualifiedCall` branch is right, and I kept it as written. The docstring edit is right too: the old "Cannot occur (rejected at check time)" claim for `QualifiedCall` was simply false, and you noticed.

## What the review found underneath it

Walking an effect-op argument does more than record its `E501`. `translate_expr` on a `FnCall` argument reaches `_translate_call_with_info`, which *assumes* the callee's `ensures` with a bare `self.solver.add(z3_post)` — and that lands on the solver's **base** assertion stack. `check_valid` folds `_path_conditions` around the *goal* only, so a fact learned inside an `if` outlives the branch and becomes an unconditional fact about the **caller's own slots** (`_build_callee_env` binds the callee's parameters to the caller's terms).

The escaped fact is circular:

```vera
private fn dec5(@Nat -> @Nat)
  requires(@Nat.0 >= 5)
  ensures(@Nat.result == @Nat.0 - 5)
  effects(pure)
{ @Nat.0 - 5 }

public fn main(@Nat -> @Nat)
  requires(true)
  ensures(@Nat.0 >= 5)          -- FALSE for input 0
  effects(<Slp>)
{ if @Nat.0 >= 5 then { Slp.sleep(dec5(@Nat.0)) } else { () }; @Nat.0 }
```

`ensures(@Nat.result == @Nat.0 - 5)` injects `ret == @Nat.0 - 5`. With `@Nat`'s implicit `ret >= 0`, Z3 derives `@Nat.0 >= 5` **globally** — the very precondition the branch guard existed to establish. So:

| | `main` | your branch |
|---|---|---|
| `vera verify` | `ok=false`, `E500` | `ok=true`, `6 verified (Tier 1)`, no diagnostics |

A program `main` correctly rejects, the branch silently accepts. Routing effect-op arguments into the call translator is what exposed it — the mechanism was already there, but nothing had reached it from this direction.

Three things follow, and the third is the one I want you to see:

1. **The same leak exists on `main` today with no effect operation anywhere.** Put `dec5` in a plain `let` inside the same `if`, and `vera verify` prints `OK / 6 verified (Tier 1)` while `vera run main -- 0` traps on the postcondition it just "proved". Filed as [#957](https://github.com/aallan/vera/issues/957) and fixed in this PR.

2. **Vacuity.** Two calls in mutually-exclusive arms inject contradictory facts about the same caller slot. The base solver goes **UNSAT**, `check_valid`'s refutation query returns `unsat` for *every* goal, and every obligation discharges vacuously.

3. **Which means the bug was silently eating your own diagnostic.** In a program whose earlier arms make the solver UNSAT, the `E501` your `QualifiedCall` branch raises is suppressed. The feature and its own regression cancelled out — which is exactly why no test caught either.

## The fix

`_translate_match` already guarded its injected facts this way:

```python
self.solver.add(z3.Implies(cond, f))
```

The call translator was the only site that didn't — `_translate_call_with_info` (`smt.py:1878-2049`) contained no reference to `_path_conditions` at all. So:

```python
def _guard_fact(self, fact: z3.ExprRef) -> z3.ExprRef:
    if self._path_conditions:
        return z3.Implies(z3.And(*self._path_conditions), fact)
    return fact
```

applied at both `solver.add` sites (the callee `ensures`, and a refined return's predicate).

My first attempt was `solver.push()` / `pop()` around your argument walk. It was sound but lossy — it discarded facts that were unconditionally true, and it couldn't reach the pre-existing statement-position leak at all. The review's guard keeps each fact exactly as strong as the path that earned it. Same soundness, strictly more precision, and it's the idiom already in the file.

Validation, each run on both sides:

- the guarded effect-op program: `ok=true` → `ok=false` + `E500`
- the pure statement-position program (pre-existing): `ok=true` → `ok=false` + `E500`
- a *legitimately* branch-guarded call (`if @Nat.0 >= 1 then { IO.print(need_pos(@Nat.0)) }`): still `ok=true`, no over-rejection
- your `#776` repro: `E501` preserved
- a nested `show(mag(x))` still discharges `show`'s precondition from `mag`'s postcondition — precision kept
- 143 conformance programs + 37 examples: `tier1 1464 → 1464`, `tier3 594 → 594`, zero diagnostic changes
- full suite green

## Small things

- **Your test discriminates**, which is more than I can say for a couple of mine. Reverting `smt.py` to `main` turns it red. Two mutants did survive it: walking only `expr.args[0]` (your `IO.print` has one argument) and returning an argument's term instead of `None`. The second is unsound — an effect operation's result is the *handler's* choice, so `ensures(@Int.result == @Int.0)` for `Counter.bump(@Int.0)` would report `verified` where it should report `tier3`. Both are pinned now.
- The `#727` dedup claim checks out: the same call visited by several passes raises one `E501`; two distinct call sites raise two.
- `KNOWN_ISSUES.md` #764's row narrowed — a call nested inside the *arguments* of an effect-op `let` value is now checked by your change, so the residual is the let-bound value itself and the statements after it.
- Filed [#958](https://github.com/aallan/vera/issues/958) separately: `spec/11-compilation.md` §11.8 says Tier-1-proved contracts are omitted from the compiled output, but `cmd_compile` never calls the verifier and `codegen/contracts.py` has no notion of tiers. That drift is the *only* reason #957 was loud rather than silent — the runtime `contract_fail` guard still fires. #958 carries an explicit dependency: §11.8 must not be implemented until #957 is fixed.

## Where this leaves your three PRs

You went after the same thing three times: places where our own self-checking machinery reports clean while something real passes through. A doc-sync test that guarded three mirrors of five. A CI gate whose skip keyed on a function's *name*. And a verifier that never visited the arguments of an effect operation.

The third one turned out to sit directly on top of a false Tier-1 proof that had been shipping in `main` — `vera verify` affirming a postcondition that traps at run time. You didn't find that bug; you found the doorway to it, and the doorway is what made it findable.

Merging. Thank you — genuinely.

Issue #987, opened 2026-07-11
#820 per-component widen guards are not emitted for imported module bodies (target-type table is single-module, span-keyed)


aallan:
## Summary

The #820 per-component `@Nat` → `@Int` widening guards (array-literal element, tuple construction/destructure component) are **not emitted for imported module bodies**. `vera verify` on the library file reports the widening as a Tier-3 runtime check, but when the function is imported and compiled into another program, that runtime check is silently absent from the persisted artifact — `vera compile`, browser bundles, and wasi-p2 components included — and the value bit-reinterprets to a negative `@Int` (`u64.MAX` → `-1`) with no trap.

## Reproduction

```vera
// widenlib.vera
public fn widen(@Nat -> @Int)
  requires(true) ensures(true) effects(pure)
{ let @Array<Int> = [@Nat.0]; @Array<Int>.0[0] }
```

```vera
// main.vera
import widenlib(widen);
public fn callit(@Nat -> @Int)
  requires(true) ensures(true) effects(pure)
{ widen(@Nat.0) }
```

```
vera verify widenlib.vera    # widen's coerce obligation: Tier 3 (runtime-guarded claim)
vera verify main.vera        # OK, clean
vera run widenlib.vera --fn widen  -- 18446744073709551615   # traps (guard fires — correct)
vera run main.vera     --fn callit -- 18446744073709551615   # prints -1 silently (guard absent)
```

WAT evidence: `vera compile --wat widenlib.vera` emits the widen guard; compiling `main.vera` (same body, imported) emits one fewer `unreachable` — the guard is gone. The tuple construction/destructure sites behave identically.

## Root cause

The checker's `expr_target_types` side-table — the target-type source the #820 guards need — is computed for the **top-level program only** and keyed by bare span (`line, col, end_line, end_col`) with **no file identity**. Imported bodies compiled via `resolved_modules` have no entries, so codegen's target recovery finds nothing and skips the guard. The verifier's `_target_type_of` recovers targets structurally per-module, so its Tier-3 classification is unaffected — producing the broken promise.

[...]

PR #1007, opened 2026-07-11
fix(codegen): run the ability-op rewrite over imported function bodies

### LINKED ISSUE #992: Imported function bodies never get the ability-op rewrite — eq/compare in an imported fn fails vera compile with unknown-func (check/verify green)
## Summary

Imported function bodies never receive the Pass-1.6 ability-op canonicalization: `eq(...)`/`compare(...)` in ANY imported function — top-level, direct `where`-helper, or nested helper — stays a raw call that codegen cannot lower, so the body is skipped and `vera compile` on the importing program fails with an internal unknown-func error. `vera check` and `vera verify` are green.

## Reproduction (confirmed at the PR #989 head)

A library whose function body uses `eq`:

```vera
-- lib.vera
public fn iseq(@Int -> @Bool)
{ eq(@Int.0, 5) }
```

```vera
-- main.vera
import lib(iseq);
public fn main(@Unit -> @Bool)
{ iseq(5) }
```

`vera check`/`verify` pass; `vera compile main.vera` fails `unknown func: failed to find name $iseq`. The same holds when the `eq` sits in an imported function's direct `where`-helper or a nested helper (three probes: top-level, child, grandchild — all fail identically). Native operators (`==`, `+`) in imported bodies are unaffected.

## Root cause

`_rewrite_ability_ops` (vera/codegen/core.py) rewrites `program.declarations` + monomorphized clones only; Pass 2.5 compiles `self._imported_fn_decls` directly, unrewritten. The same-file path rewrites `eq`/`compare` into `BinaryExpr`/if-chains (recursively through the helper tree since PR #989); the imported path never runs that pass at all.

## Notes

- Pre-existing at base (`3efad78`) — PR #989's recursion fixes did not introduce or change this; its cross-module emission fix uses native operators and is unaffected.
- Same check-green/compile-fail class as the fixed #978 and the open #990: loud failure, no silent miscompile, no false Tier-1.
- Fix direction: run the ability-op rewrite over `_imported_fn_decls` (the full helper tree) before Pass 2.5, sharing the existing recursive `_rewrite_where_fns`.

Found by the PR #989 review round (the FIX-B symmetry probe: does the imported path apply the rewrite the local path applies?).

aallan:
## Summary

Fixes #992 — the Pass-1.6 ability-op rewrite (`eq(...)` → `BinaryExpr`, `compare(...)` → if-chain) ran over the top-level program's declarations and the mono clones but never over `_imported_fn_decls`/`_shadowed_module_fns`, which compile directly in Pass 2.5/2.6. An `eq` anywhere in an imported body (top-level fn, where-helper, or nested grandchild) stayed a raw `FnCall` codegen cannot lower — the body was dropped and the importer's call dangled at WAT assembly (`unknown func`) on a check-green, verify-green pair.

## The fix

The rewrite now runs over every imported entry, via a `_rewrite_fn_ability_ops` helper factored from (and shared with) the mono-clone loop. Entries are the flattened where-tree — each helper is its own entry — and the rewrite is idempotent, so a parent's carried `where_fns` subtree cannot double-transform anything.

## Evidence

- RED-first: all five shapes failed with the exact `unknown func` diagnostics (`$check`/`$inner`/`$grandchild`) before the fix.
- `tests/test_xmod_ability_ops_992.py`: `eq` at all three nesting depths, `compare` (the other AST-rewritten ability op, matched on `Ordering`), and the shadowed (`mod$…`, Pass 2.6) door — each runs end-to-end with true/false probes on both sides of the boundary.
- Drop-the-rewrite mutant: killed (all five RED).
- Full gate green (pre-commit: suite, mypy, 157 conformance, 37 examples, doc gates).

Light-ceremony PR per the tiered-review decision (S-fix, fully grounded by the mono/import cluster report).

Closes #992

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `eq(...)` and `compare(...)` operations inside imported function bodies, including nested and shadowed-module cases.
  * Improved runtime trap diagnostics with clearer labels, kind-specific guidance, and Vera source backtraces.
  * Improved constant WASM stack handling for tail-recursive iteration.
* **New Tests**
  * Added a regression test covering imported-body ability-op rewriting for `eq(...)`/`compare(...)`.
* **Documentation**
  * Updated project status and testing totals to 7,347 tests; refreshed known-bugs listings and related entries.


aallan:
## Review record (light ceremony)

One code-reviewer agent + targeted self-probes. The fix itself held: registry-rebuild identity checked against every consumer (the generic registries hold separate decl objects and are rewritten via the mono-clone loop; intra-module renames are string-keyed; spans are preserved by the rewrite so the #987 per-module tables key correctly — the xmod widening suite through the door stays green), Pass ordering confirmed (both registries fully populated at module registration, nothing reads the unrewritten decls before Pass 1.6), and two extra door shapes probed green: `eq` over a module's own ADT (structural equality 1/0) and `eq` inside an imported **contract** (`ensures(eq(...))` compiles and the runtime postcondition passes).

**One pre-existing orthogonal bug found by the probes and filed as #1008**: an imported fn constructing its module's own ADT drops (`unknown constructor` → skip → dangling `unknown func`) unless the importer names the TYPE in its import list — constructor layouts are registered per the importer's filter. Verified at the `2743aa1` base in a pinned worktree (identical failure) and confirmed no-`eq`-involved with a control probe. Tracker rows ride this PR; the type-import workaround is in SKILL.

Issue #1015, opened 2026-07-11
Imported module where-helpers not hoisted — same-named non-generic helpers across imported fns silently run the wrong body


aallan:
## Summary

The #991 fix (PR #1013) hoists the **main program's** non-generic where-helpers to parent-qualified names before codegen, but **imported module ASTs are not hoisted** — imported function bodies compile in Pass 2.5/2.6 using bare helper names against a flat, first-seen-wins namespace. Two imported functions each carrying a same-named non-generic where-helper silently resolve to the first one's helper body: check-green, wrong value, exit 0.

## Reproduction

`lib.vera`:
```vera
public fn a(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  leaf(@Int.0)
}
where {
  fn leaf(@Int -> @Int)
    requires(true) ensures(true) effects(pure)
  { @Int.0 + 1 }
}

public fn b(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  leaf(@Int.0)
}
where {
  fn leaf(@Int -> @Int)
    requires(true) ensures(true) effects(pure)
  { @Int.0 + 2 }
}
```

`main.vera`:
```vera
import lib(a, b);

public fn main(@Unit -> @Int)
  requires(true) ensures(true) effects(pure)
{
  a(0) * 100 + b(0)
}
```

`vera run` → **101** (b's `leaf` resolved to a's +1 body). Correct: **102** (a=1, b=2).

## Root cause (located)

`_hoist_nongeneric_where_helpers` (vera/codegen/core.py) is applied only to the top-level `program`, not to each `ResolvedModule.program` before `_register_modules` collects and compiles imported bodies. Imported non-generic helpers therefore keep bare names and collide first-seen-wins.

## Provenance & classification

Found by the PR #1013 review (CodeRabbit, at head 8f0d2aa); coordinator-confirmed **pre-existing** — the same shape returns 101 at the integration base `1fd4043` (imported-body hoisting never existed). Distinct code path from #991's main-program fix. Same import-door family as #999/#1000/#1008/#1012/#1014 and the natural place to fix it is the import-door sprint PR: apply the hoist to each resolved module's program before its Pass-2.5/2.6 registration, threading the parent-qualified names the same way the main program does (mind the #987/#997 per-module span-table keying).

Workaround: give where-helpers program-unique names across imported modules until this lands.

Issue #1014, opened 2026-07-11
Nested generic where-helpers keyed flat, first-seen-wins — same-named forall helpers under different parents run the wrong body


aallan:
## Summary

`collect_nested_generic_decls` (`vera/monomorphize.py`) keys nested `forall` where-helpers by bare name, **flat and first-seen-wins** — two same-named generic helpers under *different* parents both resolve to the first one's clones. The second parent's call silently runs the first parent's body: check-green, verify-consistent, wrong value, exit 0.

## Reproduction

```vera
public fn a(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  g(@Int.0)
}
where {
  forall<T> fn g(@T -> @Int)
    requires(true) ensures(true) effects(pure)
  { 1 }
}

public fn b(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  g(@Int.0)
}
where {
  forall<T> fn g(@T -> @Int)
    requires(true) ensures(true) effects(pure)
  { 2 }
}

public fn main(@Unit -> @Int)
  requires(true) ensures(true) effects(pure)
{
  a(0) * 10 + b(0)
}
```

`vera run` → **11** (b's call ran a's body). Correct: 12.

## Root cause (located)

The nested-generic collector keys by bare helper name with first-seen-wins, so both parents' bare `g` calls monomorphize against the first-collected declaration. This is the **generic-path analogue of #991** (whose non-generic half is fixed by PR #1013 via parent-qualified hoisting + lexical call rewriting): the same parent-qualified keying / lexical scoping is needed for the nested-generic clone bases.

## Provenance

Reported by the PR #1013 fix round while working the shadow-aware rewrite; reproduced independently by the coordinator. Same family as #1012 (generic-path aunt-call scope gap) and grounds with the #999/#1000/#1002 mono cluster.

PR #1026, opened 2026-07-13
Harvest private generic module helpers

### LINKED ISSUE #1000: Private module generic reached transitively from an exported generic is never harvested (unknown func at compile)
## Summary

A **private** `forall<T>` function in a module, called by the module's **public** generic, is never harvested by `_register_modules` — so when the importer instantiates the public generic, the clone's body call to the private sibling finds no clone base (`fn_name not in generic_decls` in the transitive scan) and lowers to a dangling name. `vera check` accepts, `vera verify` is green, `vera run`/`compile` fails WAT assembly with `unknown func: failed to find name \$inner`.

The harvest in `vera/codegen/modules.py` deliberately skips private module generics on the grounds that "a private module generic is not callable from the importer" — true for direct calls, but it IS reachable transitively through the public generic's clone body, which is compiled into the importer's flat WASM module.

## Repro

`lib_priv.vera`:
```vera
private forall<T> fn inner(@T -> @T)
  requires(true) ensures(@T.result == @T.0) effects(pure)
{ @T.0 }
public forall<T> fn outer(@T -> @T)
  requires(true) ensures(@T.result == @T.0) effects(pure)
{ inner(@T.0) }
```

`main.vera`:
```vera
import lib_priv(outer);
public fn main(@Unit -> @Int)
  requires(true) ensures(@Int.result == 7) effects(pure)
{ outer(7) }
```

`vera check main.vera` → OK; `vera run main.vera` → `WAT compilation failed: unknown func: failed to find name \$inner` (exit 1).

## Provenance

Pre-existing: reproduced identically at `release/v0.1.4` (75f1ff4), before and after the #990/#998 fix branch. Found by the #990 grounding probes.

## Fix direction

Harvest private module generics as transitive-only clone bases — they must be available to the clone-body transitive scan (and compiled with their module's tables per #998), WITHOUT joining the importer-facing bare-name namespace (a private fn must stay uncallable from the importer; visibility is enforced by the checker either way, but the codegen base map must not let an importer-side bare call accidentally resolve). Grounds with the mono/import cluster.

hasnainkhatri87:
## Summary

Harvest private generic helpers from imported modules for transitive monomorphization.

An imported public generic can call a private generic helper. The helper is not visible to the importing program, but its concrete clone still must be emitted into the flattened WASM module.

Closes #1000

## Implementation

- Include private imported `forall` declarations in the code-generation generic registry.
- Preserve existing public/import-filter and shadowing behavior.
- Add a regression test covering a private `inner` helper called by public generic `outer`.

## Validation

- `pytest tests/test_codegen_modules.py tests/test_verifier_calls_modules.py -q`
  - 199 passed, 3 skipped
- `ruff check vera/codegen/modules.py tests/test_codegen_modules.py`
  - passed
- Issue reproduction using `vera run`
  - passed; program returned `7`
- `git diff --check`
  - passed

Mypy was also attempted but reported two pre-existing unused-ignore errors in untouched `vera/codegen/api.py` at lines 646 and 652.

## Checklist

- [x] Focused change
- [x] Regression test added
- [x] Existing visibility rules remain unchanged
- [x] No generated files or dependency changes
- [x] No secrets or private data added

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cross-module generics so imported public generics correctly harvest and compile private generic helpers, ensuring correct runtime behaviour.
  * Improved importer-side verification so invalid/private helper contracts reached transitively are detected and reported.
* **Tests**
  * Added regression coverage for cross-module generic monomorphisation, including a case that must fail verification.
* **Documentation**
  * Updated the Unreleased changelog entry to reflect the fix.


aallan (commented):
Thanks for the fast iterations here @hasnainkhatri87 — solid progress. Re-checked against `f4f1a77` locally: the verifier lockstep is correct (a lying private generic reached through an imported public one is now caught at verify with `E500`), `test_private_generic_with_lying_contract_is_caught_at_verify` pins it, and the CHANGELOG/lint is green. The core #1000 harvest works.

One problem remains, and it sits deeper than the harvest: **private generics are keyed by bare name on both the codegen and verifier sides**, so a name collision resolves to the wrong body.

**1. Local collision → silent miscompile.** An importer with its own local `dup`, importing a public generic that transitively calls a private `dup`:

`lib.vera`
```vera
private forall<T> fn dup(@T -> @Int) requires(true) ensures(true) effects(pure) { 0 }
public forall<T> fn caller(@T -> @Int) requires(true) ensures(true) effects(pure) { dup(@T.0) }
```
`main.vera`
```vera
import lib(caller);
private fn dup(@Int -> @Int) requires(true) ensures(@Int.result == @Int.0 + 100) effects(pure) { @Int.0 + 100 }
public fn main(@Unit -> @Int) requires(true) ensures(true) effects(pure) { dup(5) }
```
```
vera run main.vera   →   0      (should be 105)
```

`main`'s local `dup(5)` compiles to call the imported private `dup$Int` — check-clean, verify-clean, no trap. `E608` only guards cross-module collisions (`fn_provenance` is imported-only), not local-vs-import.

**2. Verifier keying** (CodeRabbit's out-of-diff point): the verifier's `_imported_generic_decls` uses the same bare-name `setdefault`, so two modules' same-named private generics collapse to one `FnDecl` and the other's contract goes unverified.

Both share a root: private-generic routing needs to be module-qualified *and* still reroute the transitive call — dropping the `and is_public` gate alone fixes the local call but breaks the transitive one. That's squarely in the mono/import name-resolution family we're actively reworking (#1014 / #1015 and neighbours), so we're going to fold #1000 into that internal pass and carry it forward from here rather than have you chase the collision handling on your own.

Really appreciate you surfacing this — and the verifier-lockstep piece was the right fix. I'll link the follow-up when it lands.

aallan:
The in-house replacement for this fix is now up as PR #1029 (the full mono/import-door cluster, including the module-qualified private-generic keying on both the codegen and verifier sides that the collision cases here required). Thanks again for the initial approach and the verifier-lockstep direction — credited in the PR. This draft will be closed when #1029 merges.

aallan:
Superseded by #1029, now merged — it fixes the private-generic transitive harvest (#1000) together with the six sibling mono/import-door bugs, using module-qualified keying (`mod$<path>$<name>`) to handle the local-vs-import collision cases in both directions plus the cross-module verifier collapse. Thanks again for the initial harvest approach and the verifier-lockstep direction in this draft.

Issue #1017, opened 2026-07-12
Verifier: @Int->@Nat narrowing into an apply_fn closure formal is not obligated (false Tier-1)


aallan:
## Summary

The verifier's `apply_fn` special case obligates the `@Nat -> @Int` **widening** direction at closure formals ([#820](https://github.com/aallan/vera/issues/820), `vera/verifier.py` ~L3208), but not the `@Int -> @Nat` **narrowing** dual. A provably-negative `@Int` passed into a closure's `@Nat` formal via `apply_fn` verifies **clean** with no E503 obligation — a false Tier-1.

## Reproduction

```vera
type NatToInt = fn(Nat -> Int) effects(pure);

private fn f(@NatToInt -> @Int)
  requires(true) ensures(true) effects(pure)
{
  apply_fn(@NatToInt.0, 0 - 5)
}
```

`vera check` — clean. `vera verify` — **ok: true, no nat_bind obligation** (should be E503: `0 - 5` is provably negative narrowing into a `@Nat` formal).

## Root cause (located)

In `vera/verifier.py`'s `apply_fn` branch, the formals loop over `_closure_arg_param_types` checks only `_is_int_type(formal) and _result_is_nat(arg)` (the widening obligation). It needs the narrowing dual: `_nat_binding_target(arg, formal) and _narrows_into_nat(arg)` → `_check_nat_binding_obligation`, mirroring the generic call-argument path, with the matching `call_indirect` runtime guard (the #754 runtime-guard family covers whether the guard is emitted or deferred).

## Provenance & classification

Found by the PR #1016 (v0.1.4 release) CodeRabbit review of the accumulated `release/v0.1.4...main` diff. **Pre-existing** — the `apply_fn` narrowing site was never obligated (the #820 work added only the widening direction at this site). Verification-completeness gap; groups with ROADMAP Stage 19's narrowing-obligation rows and the #754/#757 runtime-guard family. Deferred from v0.1.4 (a docs/version-only release PR) to the v0.1.5 bug sprint.

aallan:
[...]

Issue #984, opened 2026-07-10
Closure @Int -> @Nat return narrowing is neither obligated nor runtime-guarded (the #758 hole through _compile_lifted_closure)


aallan:
## Summary

An `@Int -> @Nat` narrowing at a **lifted closure's** return position is neither statically obligated nor runtime-guarded — the exact #758 soundness hole, reachable through the closure compilation path that the #758 fix (top-level functions and where-helpers via `_compile_fn`) does not cover.

```vera
type IntToNat = fn(Int -> Nat) effects(pure);

private fn make_narrow(@Int -> @IntToNat)
  requires(true) ensures(true) effects(pure)
{
  fn(@Int -> @Nat) effects(pure) { @Int.0 }
}

public fn go(@Int -> @Nat)
  requires(true) ensures(true) effects(pure)
{
  let @IntToNat = make_narrow(@Int.0);
  apply_fn(@IntToNat.0, @Int.0)
}
```

- `vera verify` → OK, Tier 1 (no E503 on the closure's return narrowing)
- compiled WAT: the lifted closure body is a bare `local.get` — no return guard
- `vera run --fn go -- -5` → `-5` through the declared `@Nat` slot, no trap

The identical narrowing as a top-level function or where-helper is E503-obligated and runtime-trapped since #758's fix. Pre-existing at that fix's base (not a regression); same family as #779's closure/quantifier/handler-body obligation-walker gap — the verifier's obligation walkers and codegen's function-boundary guards both stop at `_compile_lifted_closure`. Found by the PR #983 adversarial panel's TCO/emission-consistency lens.

aallan:
Root cause: the closure body's return leaf was outside both halves of the narrowing machinery — the verifier's narrowing walk treats `AnonFn` as a terminal (correctly, per the closure-opacity rationale) but had no shallow obligation for the narrowing direction, and `_compile_lifted_closure` emitted no return guard — so `fn(@Int -> @Nat) { @Int.0 }` applied to a negative returned it through the `@Nat` slot silently on a verify-clean program, while top-level functions and where-helpers were covered by #758.

[...]

Issue #820, opened 2026-06-29
@Nat → @Int widening: remaining deferred coercion sites (follow-up to #813)


aallan:
## Summary

The `@Nat → @Int` widening soundness work ([#813](https://github.com/aallan/vera/issues/813)) guards every site where code generation can statically determine the source is `@Nat` — return, `let`, call argument, concrete `@Int` constructor field, ADT sub-pattern extraction, match binding — and its follow-up in **#821** added two more: the explicit `nat_to_int` built-in and heterogeneous `if`/`match` arms whose alternative is a **non-negative literal**. This issue tracks the sites that remain deferred.

### Already fixed (do not re-do)

- **`nat_to_int(@Nat)` built-in** (was section D below) — its declared `@Int` return had masked the `@Nat` source, so `vera verify` falsely proved `ensures(@Int.result >= 0)` while `vera run -- 18446744073709551615` returned `-1`. Fixed in #821; it now carries the `nat_to_int_coerce` obligation + the boundary trap guard, which caught real unguarded widenings in `json.vera` / `life.vera`.
- **Heterogeneous `if`/`match` with a non-negative-literal alternative** (`if c then { @Nat.0 } else { 0 }`) — part of section C below. Fixed in #821 by treating a non-negative literal arm as `@Nat`-compatible, so the whole expression stays `@Nat` and the single boundary guard covers the real `@Nat` arm.

### Moved to the narrowing-dual issues

Three deferred widening sites share the **exact** blocker of their `@Int → @Nat` narrowing duals (code generation has no per-component target-type metadata), so they are tracked alongside the narrowing direction rather than duplicated here:

- **Effect-operation argument** → [#754](https://github.com/aallan/vera/issues/754)
- **Generic-instantiated `@Int` field** (`Some(@Nat.0)` into `Option<Int>`) → [#757](https://github.com/aallan/vera/issues/757)
- **Tuple construction / destructure component** → [#758](https://github.com/aallan/vera/issues/758)

## Remaining scope (this issue)

The widening sites with no existing narrowing-dual home, all blocked on the same missing per-component target-type metadata in code generation:

- **Array-literal element** — `let @Array<Int> = [@Nat.0]`. The literal is typed by its element *values* (source) and the `let` is a pointer-pair copy, so the element coercion has no guard site (`_infer_array_element_type` returns the source type). Currently disclosed as an unguarded **E531** warning. (The `@Int → @Nat` array-element narrowing is not separately tracked, so it belongs here too.)
[...]

PR #1033, opened 2026-07-16
fix: obligate and runtime-guard refined closure returns (#1032)

### LINKED ISSUE #1032: Refined-predicate closure RETURN is not obligated or guarded (false Tier-1) — the return dual of #1024
## Summary

The refined-predicate dual of #1024 on the closure **return** boundary: a closure declared `fn(@Int -> @Pos)` (where `type Pos = { @Nat | @Nat.0 > 0 };`) whose body returns its raw `@Int` argument verifies **clean** and, applied to `-5` via `apply_fn`, returns `-5` through the refined slot with no trap.

Found by PR #1030-era review work on #1024 (the formal-side fix): the closure-return path has the #984 `@Nat` narrowing guard (`>= 0` base, per narrowing leaf) and the #820 widening guard, but no refinement-predicate arm — the same three-arm asymmetry #1024 closed on the argument side.

## Repro

```vera
type Pos = { @Nat | @Nat.0 > 0 };
type IntToPos = fn(Int -> Pos) effects(pure);

private fn f(@IntToPos, @Int -> @Pos)
  requires(true)
  ensures(true)
  effects(pure)
{
  apply_fn(@IntToPos.0, @Int.0)
}

public fn main(@Unit -> @Int)
  requires(true)
  ensures(true)
  effects(pure)
{
  let @Pos = f(fn(@Int -> @Pos) effects(pure) { @Int.0 }, 0 - 5);
  nat_to_int(@Pos.0)
}
```

`vera verify` — clean.  `vera run` — returns `-5` through the `@Pos` slot (both the `>= 0` base and the strict `> 0` violated; the #984 leaf guard is excluded when the return type is a refinement — the refinement boundary was assumed to own the check, but no refinement-boundary guard exists on the closure return).

## Fix direction

Mirror #1024's two sides at the closure RETURN: a verifier obligation on the `AnonFn` return (shallow-syntactic tier3-guarded, matching the #820/#984 closure-return arms — closure bodies stay opaque to SMT), and a codegen refinement check on the return value in `_compile_lifted_closure` (the return-side analogue of the #1024 entry guards), keyed via `_refinement_guard_parts(anon_fn.return_type)`.  The named-function refined return already has both halves — port that convention.

JSap0914:
## What

A closure whose declared return is a refinement `{ @Base | P }` is now obligated by the verifier and guarded at run time, closing the return-side dual of #1017 (the closure-argument narrowing) and the false Tier-1 reported in #1032.

## Why

Given `Pos = { @Nat | @Nat.0 > 0 }`, the closure `fn(@Int -> @Pos) { @Int.0 }` returns its raw `@Int` argument through the refined slot. Before this change, a program applying it verified **clean** and, at run time, returned the violating value with **no trap**:

```vera
type Pos = { @Nat | @Nat.0 > 0 };
type IntToPos = fn(Int -> Pos) effects(pure);

private fn pass_through(@IntToPos, @Int -> @Int)
{ let @Pos = apply_fn(@IntToPos.0, @Int.0); nat_to_int(@Pos.0) }

public fn main(@Unit -> @Int)
{ pass_through(fn(@Int -> @Pos) effects(pure) { @Int.0 }, 0 - 5) }
```

`vera verify` → `4 verified (Tier 1)` (no obligation); `vera run` → `-5`.

`pass_through` deliberately returns `@Int`, not `@Pos`, so its own named-return guard cannot mask the hole. When the consumer *does* return `@Pos`, the named-function refined-return guard (#746) catches it — which is exactly why the closure boundary was trusted but never enforced: the refinement was assumed established where it was produced (the closure return), but nothing checked it there.

## Change

- **Verifier** (`vera/verifier.py`): the `AnonFn` walk gains a refinement arm beside the existing #820 widening and #984 narrowing return arms. The closure body is opaque to SMT, so the refined return is obligated shallow-syntactically — always a guarded `tier3`, never a Tier 1 — via a new `_return_narrows_into_refined` join walk (the refinement analogue of `_return_narrows_into_nat`).
- **Codegen** (`vera/codegen/closures.py`): `_compile_lifted_closure` now saves the body result, runs the refinement predicate guard, and pushes it back — the closure analogue of the named-function `_compile_postconditions` refined-return guard, reusing `_refinement_guard_parts` + `_emit_refinement_check`. Scalar returns only; a refined pair (String/Array) return keeps its current behavior.

After the change: `vera verify` → `1 runtime checks (Tier 3)`; `vera run` on `-5` (and on `0`) traps at the closure boundary; a satisfying value passes unchanged.

## Tests

`tests/test_refined_closure_return_differential.py` — a verifier↔codegen behavioural differential mirroring the #984 narrowing dual:

- The closure-return `refine_bind` obligation is exactly `["tier3"]` (written first; fails as `[]` — a false Tier 1 — without the fix).
- A strictly-negative value (violates `>= 0` and `> 0`) **and** a zero (violates only the strict `> 0`) both trap at the closure boundary — proving the full predicate is enforced, not just the `@Nat` base.
- A satisfying value passes; a non-refined `@Int -> @Int` closure return carries no obligation and no guard.

```
uv run pytest tests/test_refined_closure_return_differential.py   # 3 passed
uv run pytest                                                     # 7382 passed, 81 skipped
uv run pre-commit run --files <changed>                          # all hooks pass (ruff, mypy, pytest, conformance, doc-counts, walker coverage, diagnostic fields)
```

Fixes #1032

## Summary by CodeRabbit

- **Bug Fixes**
  - Refined return types on closures are now both verifier-obligated and runtime-guarded.
  - Closure return values that violate refinement requirements now trap at the closure boundary (including refined `i32` and refined string/pair-return shapes).
  - Non-refined closure returns still pass through unchanged without additional checks.

- **Tests**
  - Added a verifier↔codegen differential test suite for refined closure returns (including pair-based closure returns) and updated expectations for obligations and traps.

- **Documentation**
  - Updated the changelog, testing documentation, and project test metrics.


JSap0914:
Thanks for the review. Addressed the P1 in the follow-up commit: the refined closure-return guard now also covers `i32_pair` bases (`@String` / `@Array`), mirroring the named-function `_compile_postconditions` i32_pair path — it saves and reloads the `(ptr, len)` halves and checks the predicate on the ptr. This makes the verifier's `guarded=not _is_unit_refinement(...)` flag honest for every non-`@Unit` refined closure return (no `tier3_runtime` promise over an unguarded passthrough). Added a `{ @String | string_length > 0 }` differential case: an empty string traps at the closure boundary, a non-empty one passes; verified red without the pair guard.

Re the P2 (local allocated before confirming `_emit_refinement_check` can emit the guard): this matches the existing named-function guard in `_compile_postconditions`, where the result/ptr/len locals are likewise allocated before the `guard is None` check — the helper needs the local index to build the guard, so the allocation necessarily precedes it. Kept consistent with that convention.

aallan:
Thank you for this, @JSap0914 — and welcome! This is a genuinely well-made fix: the `_return_narrows_into_refined` join-walk is exactly the right shape, the save-guard-push epilogue mirrors the named-function mechanism correctly, your second commit closing the `i32_pair` return gap was the right call, and the repro's insight of having the consumer return `@Int` so the named-function guard can't mask the closure hole is sharper than the one on the issue.

An apology is owed: we had our own fix for #1032 in flight — implemented and validated locally as part of the active bug-burndown sprint, but not yet pushed to GitHub when you opened this, so there was no way for you to know. That's on us; it's now up as #1034 (together with the #1024 argument-side dual it was built alongside), and we're going to land that one since it's integrated with the argument-side machinery and tests in the same differential file.

One piece of technical feedback you may find useful, since your implementation is so close: the verifier arm's **ordering** matters more than it looks. `_is_int_type` (and `_is_nat_type`) deliberately match a `RefinedType` over their base — so with the refined arm placed as an `elif` after the #820 widen arm, a closure returning `{ @Int | P }` with a `@Nat`-producing body gets claimed by the widen arm and the refinement predicate is never obligated (your codegen guard still traps at runtime, so it's an obligation-kind mislabel rather than a false Tier-1 — but the obligation stream is meant to describe exactly what codegen guards). #1034 runs the refined arm first for this reason, and also re-derives the `test_refined_return_single_guard_no_double` pin so it counts the refinement guard explicitly rather than passing by omission.

If you'd like to contribute — and on this evidence we'd be glad to have you — some open work that's well clear of the current sprint (nobody will race you on these):

- **#610** — interpret a minimal ANSI escape subset in the browser runtime (`runtime.mjs`): self-contained JS, no compiler-core surface, crisp acceptance criteria.
- **#770** — non-blocking / timed stdin so a program can wait on a key press and a clock at once: host-runtime work with real design latitude.
- **#686** — implement `data invariant(...)` clauses per spec §2.6/§6.2.3: a bigger, well-specified checker+verifier feature if you want something meatier.

A note on conventions if you pick one up: the repo's workflow is test-first (RED before the fix, see CLAUDE.md), and CHANGELOG/docs ride in the same PR. Closing this in favour of #1034 — thanks again for the quality of the work here.

Issue #1039, opened 2026-07-16
let bindings of pair-represented Future payloads (let @Future<String>) are check-green but E602-skip


aallan:
## Summary

`let @Future<String> = async("hello");` — and by extension any `Future<T>` whose payload is pair-represented (String/Array) bound via `let` — is check+verify green but E602-skips the enclosing function ("let binding type 'Future<String>' has no WASM representation"). The direct (non-alias) spelling; unrelated to the #1031/#1037 alias canonicalization (verified: identical with and without that fix). Found by PR #1038's family mapping: it blocks the `await(Future<String>)` match-scrutinee shape upstream, with no match required to reproduce.

## Root cause direction

`Future<T>` is representation-transparent (#841), so `Future<String>` is an `i32_pair` — but the let-binding translator's pair handling appears keyed to the known pair type names rather than the canonical WASM type, so the transparent wrapper falls to the no-representation skip. Same check-green→E602 family as the fixed #1006/#1031/#1037/#1038.

## Repro

```vera
public fn go(@Unit -> @Int)
  requires(true)
  ensures(true)
  effects(<Async>)
{
  let @Future<String> = async("hello");
  let @String = await(@Future<String>.0);
  string_length(@String.0)
}
```

`vera check`/`verify` green; `vera run` → E602 skip → unknown func.

aallan:
Root cause: the let path's pair predicate (`_is_pair_type_name`) keyed on the literal `String`/`Array` names, so a pair-represented Future payload (`Future<String>`, `Future<Array<T>>`) fell to the scalar branch — where the scalar mapper recursed through the transparent wrapper to a pair inner and returned no WASM type, E602-skipping the function check-green.

Fix: the predicate is now Future-transparent (strip `Future<...>`, recurse on the payload), mirroring the scalar mapper's existing Future arm — pair payloads bind and read two locals exactly like a bare `String`/`Array` let; scalar/pointer payloads stay single locals; `Future<Unit>` still skips loudly.

Fix in #1041 — closes on merge.

Issue #1037, opened 2026-07-16
_slot_name_to_wasm_type drops alias targets' type arguments: type FI = Future<Int>; components are check-green but E602-skip


aallan:
## Summary

An alias *to* a compound representable type — `type FI = Future<Int>;` used as a destructure/match component — is check+verify green but `CodegenSkip`s the enclosing function (E602): `_slot_name_to_wasm_type` resolves aliases through the same name-only recursion #1035's review fixed in `_slot_name_erases_to_unit` (`"FI"` → `'Future'`, dropping `<Int>`), so the wrapper never resolves to its payload's WASM type and the component falls to the unrepresentable branch.

Found by PR #1035's adversarial review; verified pre-existing (differentially identical on main). Loud, never a silent erase — the #1035 canonicalizer deliberately returns False for it.

## Fix direction

Give `_slot_name_to_wasm_type`'s alias resolution the same full-compound-spelling canonicalization `_slot_name_erases_to_unit` now has (hop-by-hop with a cycle cut) — same subsystem, same helper family; fold-in candidate for the #1035 branch.

aallan:
Root cause: `_slot_name_to_wasm_type` resolved type aliases by name only — a component spelled through an alias (`type FI = Future<Int>;`) never reached the Future-transparency arm, so the alias-spelled component got no WASM type and the enclosing function E602-skipped check-green.

Fix: alias resolution now canonicalizes hop-by-hop to the full compound spelling (`_canonicalize_alias_slot_name`, cycle-cut, shared with the zero-size predicate from #1031), so alias-spelled components resolve exactly like their expansions.

Fix in #1035 — closes on merge.

PR #1040, opened 2026-07-16
fix: infer await(Future<T>) match scrutinees (#1038)

### LINKED ISSUE #1038: Match scrutinee WASM-type inference cannot type await(Future<ADT>): check-green program E602-skips
## Summary

`match await(@Future<Option<Unit>>.0) { ... }` — and the `Future<ADT>` scrutinee family generally — is check+verify green but E602-skips the enclosing function with "could not infer match scrutinee WASM type": the scrutinee inference path (`_ref_type_name_wasm_type` region) does not resolve an `await`-of-`Future<ADT>` result. Awaiting into a separate `let` first works (the component itself is preserved correctly), so this is scrutinee-position inference, not value handling.

Found by PR #1035's adversarial review; verified pre-existing (identical on main), unrelated to the zero-size guards that PR touches. Loud, not silently wrong.

## Fix direction

Teach the match-scrutinee WASM-type inference the `await` result type (unwrap `Future<T>` to T, as the let-binding path already does), or route the scrutinee through the same inference the let path uses. Same check-green→E602 family as the fixed #1006/#1031.

### LINKED ISSUE #1054: Alias to a Future (type FA = Future<Opt>) E602-skips at match-scrutinee and ctor-arg sites the direct spelling compiles
An **alias to a Future** E602-skips at the two call sites the #1038 fix covers, while the direct spelling compiles — check-green AND verify-green:

```vera
type FA = Future<Option<Int>>;

public fn f(-> @Int)
  requires(true) ensures(true) effects(<Async>)
{
  let @FA = async(Some(5));
  match await(@FA.0) {
    Some(@Int) -> @Int.0 + 1,
    None -> 0 - 7
  }
}
```

`vera check` and `vera verify` pass; `vera compile` E602-skips the scrutinee inference and then errs hard (`unknown func $scrut_opt`). The direct `match await(@Future<Option<Int>>.0)` runs (6). The same alias gap reproduces at the constructor-argument site (`data WrapI { WI(FA, Int) }`, `WI(@FA.0, 7)` skips; the direct spelling is the passing pin).

Root cause: `_ref_type_name_wasm_type` (vera/wasm/inference.py) resolves the slot name via the name-only `_resolve_base_type_name`, which drops the alias's type arguments (`FA` → bare `Future`), and its Future arm guards on AST `type_args` that an alias-spelled ref does not carry. The sibling `_slot_name_to_wasm_type` avoids this by canonicalizing through `_canonicalize_alias_slot_name` first (#1037) with a string-form Future arm. The alias analog of #1046 (`_is_pair_type_name`'s gap), one decider over — the last mapper in the family without the canonicalizer.

Loud failure (not a silent miscompile), but it breaks the #1031/#1037 alias-transparency invariant. Found by the adversarial review of PR #1040 at dae74f9, which also probe-proved the mirror fix (canonicalize + string-form strip) makes both shapes run.

aallan:
## Summary

Fixes #1038: `match await(@Future<ADT>.0) { ... }` — and the `await(Future<T>)` match-scrutinee family generally — was **check+verify green but E602-skipped** the enclosing function ("could not infer match scrutinee WASM type"). Found by PR #1035's adversarial review; verified pre-existing on main and unrelated to the zero-size guards that PR touches.

## Root cause and fix

The scrutinee inference reaches `_ref_type_name_wasm_type` (the SlotRef/ResultRef mapper in `wasm/inference.py`) through the `await` FnCall arm — and that mapper, alone among its three siblings, did not treat `Future<T>` as WASM-transparent (#841), so a `@Future<T>` slot ref inferred `None` and the function dropped. The fix adds the Future-transparency arm delegating to `_canonical_wasm_type(type_args[0])` — mirroring the let-binding path's existing mechanism (the semantic oracle: await-into-let-then-match always worked).

## Family map

| Scrutinee shape | Before | After |
|---|---|---|
| `await(Future<Option<Int>>)` | E602 skip | fixed |
| `await(Future<UserADT>)` | E602 skip | fixed |
| `await(Future<Int>)` | E602 skip | fixed |
| `await(Future<String>)` | blocked **upstream** at `let @Future<String>` — a distinct let-binding pair gap, no match required to reproduce | out of scope, filed as #1039 (PR F pending) |
| await-into-let-then-match | works | still works (pinned) |

## Validation

- **RED-first**: 6 codegen tests (`TestAwaitMatchScrutinee1038`), one per broken family member, each exercising **both** match arms with distinguishing payload-carrying values; each confirmed check-green + E602-RED before the fix. Plus the let-then-match oracle pin (green throughout).
- **Mutation**: neutering the new arm flips all 6 family tests RED while the oracle pin stays green — the tests exercise exactly this arm.
- **Mapper probe**: `Future<Int>`→i64, `Future<Bool>`→i32, `Future<String>`→i32_pair, `Future<Option<Int>>`/`Future<UserADT>`→i32, bare `Future`→None (defensive), non-Future refs unchanged.
- **Full gate green** on main's base: full pytest, mypy, ruff, 160/160 conformance, 38/38 examples, all doc/census gates including `check_e602_clean`.
- Includes the `tests/test_release.py` hermetic `GIT_*`-scrub safety pickup (content-identical to the #1034/#1035 branches; resolves cleanly whichever merges last).

## Adversarial-review round: #1054

An independent adversarial review of this PR enumerated the fixed mapper's two callers across seven expression positions (silent-wrong hunt: negative — every reachable `Future` shape is correct or loudly skipped) and confirmed one finding: an **alias to a Future** (`type FA = Future<Option<Int>>;`) E602-skipped at exactly the two sites this PR fixes, because the mapper lacked the #1037 alias canonicalization its sibling carries and its Future arm guarded on AST type args an alias-spelled ref does not have. Fixed here: canonicalize-first plus the string-form Future arm (the alias analog of #1046, closing the last mapper in the family); three tests (alias scrutinee, alias ctor-arg, alias chain), each RED as the E602 drop pre-fix, both edit sites mutation-killed independently.

Closes #1054

Closes #1038

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `await` expressions being skipped when used in `match` conditions.
  * Fixed aliases for `Future` values failing in certain mappings and constructor arguments.
  * Improved handling of `Future` values containing options, custom types, and integers.
  * Corrected match-arm selection and payload extraction for asynchronous values.

* **Tests**
  * Added regression coverage for asynchronous matching, aliases, and constructor arguments.
  * Updated reported test totals to 7,559 tests, with 7,458 passing.


aallan:
@coderabbitai full review

aallan:
## Adversarial review record

An independent reviewer attacked this PR's final implementation (head dae74f9), aimed at the code the fix newly reaches rather than the edited lines. Method: enumerated the fixed mapper's two callers (`SlotRef`/`ResultRef` arms of `_infer_expr_wasm_type`) and probed every reachable expression position with check-green differentials against `main` — match scrutinee, constructor argument, match-arm body, `apply_fn` argument, binary operand, postcondition `ResultRef`, plus edge shapes (`Future<Future<Int>>` double-wrap, refined payloads, `Future<Unit>`, the known-blocked `Future<String>` scrutinee).

**Silent-wrong hunt: negative.** Every reachable `Future` shape either produces the correct value or fails loudly; the construction and read sides agree by construction (verified end-to-end at the pinned ctor-arg site). Two positions the fix repairs beyond its original pins (match-arm body, `apply_fn` argument) were confirmed correct.

**One finding, confirmed and fixed in this PR** (#1054): an alias to a Future (`type FA = Future<Option<Int>>;`) E602-skipped at exactly the two sites this PR fixes, while the direct spelling compiled — the mapper lacked the #1037 alias canonicalization its sibling carries. The review probe-proved the mirror fix before it was written; it landed with three RED-first tests and independent mutation kills on both edit sites.

PR #1049, opened 2026-07-16
fix: zero-size field offsets — registered layouts and type-param wildcard walks (#1043, #1060)

### LINKED ISSUE #1043: Registered constructor layouts give zero-size fields 4 bytes: wildcard tag walks and structural Eq read wrong offsets
The **registered** constructor layouts (`_compute_constructor_layout`, vera/codegen/registration.py) give a declared zero-size field (`Unit`, transparent `Future<Unit>`) a 4-byte `"i32"` slot — `_resolve_field_wasm_type` falls back to `"i32"` for Unit ("shouldn't appear, safe fallback") — while **construction** (`_translate_constructor_call`, vera/wasm/data.py) lays the same field out at zero width. Every consumer that trusts `layout.field_offsets` for such a constructor reads the wrong offsets. Two confirmed check-green symptoms at 7cbc336 (`main`):

**1. Wildcard tag walk + user-ADT nested extraction** — prints **0** (expected 4242; the arm check passed via zeroed fresh-alloc memory and extraction read zeros):

```vera
private data BB { MkBox(Int), MkNot(Int) }
private data P { MkP(Unit, BB) }

private fn mkp(@Int -> @P)
{ MkP((), MkBox(@Int.0)) }

public fn f(-> @Int)
{
  let @P = mkp(4242);
  match @P.0 {
    MkP(_, MkBox(@Int)) -> @Int.0,
    MkP(_, MkNot(@Int)) -> 0 - @Int.0
  }
}
```

**2. Structural Eq** — prints **0** / false (expected true; the control without the `Unit` field prints 1):

```vera
private data BB { MkBox(Int), MkNot(Int) }
private data P { MkP(Unit, BB) }

public fn f(-> @Bool)
{
  let @P = MkP((), MkBox(7));
  let @P = MkP((), MkBox(7));
  @P.0 == @P.1
}
```

The auto-derived `$eq_` walks the registered `field_offsets` — `((4, "i32"), (8, "i32"))` — while construction stored the `BB` pointer at offset 4, so it compares a garbage word against the real pointer.

Fix direction: make the registered layout agree with construction — `_resolve_field_wasm_type` returns `"unit"` for zero-size field types, `_wasm_type_size`/`_wasm_type_align` (vera/codegen/memory.py) learn `unit -> 0 / 1` — then audit every `field_offsets` consumer (structural eq/show/hash emitters, the wildcard arm of `_sub_pattern_wasm_type`, extraction paths, calls_handlers.py:391/537, operators.py:773) for `"unit"` handling (skip the load/store; a zero-size field is equal by definition under Eq).

Sibling of the nested-tag-walk width bug (same divergence family, different consumer set).

### LINKED ISSUE #1060: Wildcard over a type-param field instantiated to Unit (Box<Unit>) silently reads following fields at wrong offsets
**Silent wrong values**: a `_` wildcard over a **type-parameter field instantiated to `Unit`** reads every following field at the wrong offset — check-green, no warnings:

```vera
private data BB { MkBox(Int), MkNot(Int) }
private data Box<T> { MkB(Int, T, Int) }

public fn f(-> @Int)
{
  let @Box<Unit> = MkB(11, (), 22);
  match @Box<Unit>.0 {
    MkB(@Int, _, @Int) -> @Int.0
  }
}
```

prints **0** (expected 22). Same for `Named<T> { N(T, String) }` as `Named<Unit>` (String read at wrong offset) and `Entry<T> { En(T, BB) }` with a nested constructor pattern (0 instead of 314). Pre-existing on `main` (proven by mutation: reverting #1049's registration fix leaves this case identically broken while regressing the declared-Unit case — the two are independent).

Root cause: construction erases an instantiated-`Unit` type-param field to zero bytes (it does not box it — WAT-verified), but the wildcard arms of `_extract_constructor_fields` (vera/wasm/data.py ~768) and `_sub_pattern_wasm_type` (~854) advance by the **registered generic** width — a type param registers as 4-byte `i32` — which is not instantiation-aware. Only the wildcard path is affected: `@Unit` bindings, `let`-destructures, and the eq/show/hash emitters all recompute from the concrete instantiation (`_resolve_field_type_for_eq` + `_concrete_field_layout`) and return correct values for the same shapes.

Alignment-masked in some layouts (a following i64/f64 re-aligns past the spurious +4, e.g. `Tag(T, Int)` as `Tagged<Unit>` is coincidentally correct), which is why it evades casual testing.

Fix direction: thread the concrete instantiation into the two wildcard walks the way the eq/show path already recomputes it. The declared-`Unit`/`Future<Unit>` spelling of the same sites is #1043 (PR #1049); this is the type-param spelling that the registered layout alone cannot know.

Found by the adversarial review of PR #1049 at cc90ba4.

### LINKED ISSUE #1070: Erases-to-Unit type ARGUMENTS spelled non-literally (Box<U>, Box<Future<Unit>>) silently break wildcard walks and Eq
**Silent wrong values**: an erases-to-Unit **type argument spelled non-literally** — `Box<U>` (`type U = Unit;`), `Box<Future<Unit>>`, `Box<FU>`, alias chains — breaks the #1060 wildcard walks and structural `Eq`, while the literal `Box<Unit>` is correct:

```vera
type U = Unit;
public data Box<T> { MkB(Int, T, Int) }

private fn mk(-> @Box<U>)
{ MkB(11, (), 22) }

public fn f(-> @Int)
{
  let @Box<U> = mk();
  match @Box<U>.0 { MkB(@Int, _, @Int) -> @Int.0 }
}
```

returns **0** (expected 22; WAT differential: the trailing `@Int` loads at offset 24 instead of 16 — past the allocation). The nested-constructor variant reads the nested tag at the shifted address (0 instead of 314). Structural `Eq` over `Box<U>` compares two equal structs unequal (0, should be 1) — a **pre-existing** sibling with the same root (it falsifies the #1060 round's "eq/show/hash were already correct" claim, which holds only for the literal spelling). show/hash loud-skip these types rather than mis-compute.

Root cause: the width classification both paths share (`_eq_field_wasm_type`, vera/wasm/operators.py) treats a field as zero-size only when its base string is literally `"Unit"` — but construction erases via the erasure-aware `_slot_name_erases_to_unit`. Declared fields dodge this because registration canonicalizes them (#1043); a **type argument** comes from the scrutinee's rendered type and is never canonicalized. One erasure-aware fix in `_eq_field_wasm_type` closes the wildcard walks (on the PR #1049 branch) and the pre-existing Eq sibling together.

Minor rider for the same round: `_later_sub_pattern_reads` counts a zero-width `@Unit` binding as a "later read," so an unrecoverable direct-call scrutinee with a trailing `@Unit` binding false-skips loudly — over-conservative, correctness-preserving.

Found by the #1060-delta adversarial review of PR #1049 (probe record in the review); merge blocker for that PR.

aallan:
## Summary

Closes both spellings of the **zero-size-field offset divergence** in the match/eq/show/hash consumers:

- **#1043** — the registered constructor layouts gave a declared zero-size field (`Unit`, transparent `Future<Unit>`, aliases) a spurious 4-byte slot while construction stores it at zero width, so wildcard tag walks matched wrong arms off zeroed memory, structural `Eq` compared garbage words (equal values unequal), `show` rendered garbage bytes, and `hash` folded them — all check/verify-green.
- **#1060** — the type-parameter spelling of the same wildcard walks: a `Box<Unit>` wildcard advanced by the registered *generic* width (4-byte i32) over a field construction erases to zero bytes, silently reading every following field at the wrong offset (0 instead of 22, no warnings). Found by this PR's adversarial review; proven independent of #1043 by mutation.

## The fixes

**#1043**: registration returns the `"unit"` sentinel for erases-to-Unit fields (size 0/align 1 in the shared maps), field type names canonicalize to `Unit` (also resolving a checker↔codegen Eq-derivability desync that spuriously E613'd `Future<Unit>`-field ADTs), and every `field_offsets` consumer gained a `"unit"` arm or is proven unable to receive one (pattern walks, `$eq_`/show/hash emitters, field loads; the builtin variadic `Tuple` stays per-call-recomputed, pinned).

**#1060**: the two wildcard walks recompute a bare type-param wildcard's width from the scrutinee's **concrete** type args (the eq/show model), threaded through both the tag-check and extraction walks with nested recursion. A direct-call scrutinee whose instantiation is unrecoverable AND whose wildcard is followed by a read **loud-skips** rather than guessing (both behaviors pinned; the recoverable SlotRef form — every corpus shape — compiles correctly; the residual completeness gap is #1065, `limitation`).

## Validation

- **RED-first**: 36 tests across two files — the wildcard/Eq/show/hash end-to-end shapes for both spellings (erased-first/last/multi positions, alignment-boundary layouts, `Future<Unit>` end-to-end), registered-layout structural pins for every spelling including alias chains, hash payload-sensitivity, the alignment-masked control that must stay green, loud-skip pins, and no-erased-field controls.
- **Mutation matrices**: nine #1043 edit sites and the #1060 instantiation-threading each reverted independently — exactly their own tests flip RED (the #1060 revert leaves all 18 #1043 tests green and vice versa, proving the two spellings independent).
- **Merged-state validation**: the #1043 diff was applied onto main in a throwaway worktree during development — all probes green there, including `Future<Unit>` end-to-end shapes.
- **Independent adversarial review** (record on this thread): no PR-introduced regression across alignment/guard-indexing/handler/hash probes; its one confirmed finding IS #1060, fixed here; a delta pass over the #1060 commit runs before merge.
- **Full pre-commit gate** on the rebased tree (current main); counts oracle-reconciled (7,605 across 123 files).

Closes #1043
Closes #1060
Closes #1070


aallan:
## Adversarial review record

An independent reviewer attacked this PR's implementation (head cc90ba4) where its 18 tests don't look: alignment interactions around erased fields (`MkP(Int, Unit, Float64)`, `MkP(Bool, Unit, Int)`, `MkP(String, Unit, BB)` — construct/extract-every-field/eq/show), the #747/#813 guard parallel-array indexing after erased fields (a misindexed guard would be a silent soundness hole — verified correctly targeted), `Exn`/`State` handler payloads over erased-field ADTs, two-level nested hash/show with payload-sensitivity, `Option<Unit>` through eq, and the `$eq_` refactor's all-unit/mixed constructor arms. ~1050 adjacent tests re-run green.

**Verdict: no PR-introduced regression; the fix is mutation-verified load-bearing.**

**One confirmed pre-existing silent-wrong adjacent to scope** (#1060): a wildcard over a **type-parameter** field instantiated to `Unit` (`Box<Unit>` matched `MkB(@Int, _, @Int)`) reads following fields at the wrong offset (0 instead of 22, no warnings) — construction erases the field but the wildcard walks trust the registered generic width. Proven independent of this PR by mutation (reverting this PR's fix leaves that case identically broken). Only the wildcard path is affected; bindings, destructures, and eq/show/hash recompute concretely and are correct. A fix round threading the concrete instantiation into the two wildcard walks is in flight on this branch.

Secondary: the test-file comment claiming `Future<Unit>` construction is blocked is stale — it works end-to-end at this head; the fix round upgrades the layout-only pins to end-to-end tests.

aallan:
## Adversarial review record — #1060 delta verdict

The dedicated pass over the #1060 commit is in: ~17 probe families over the instantiation-threading — shadowed same-base-name slots (safely keyed by full parameterized type), multi-param position mapping including reversed constructor field order, wildcards over the non-erased param, three-level nesting, the #1043+#1060 erasure composition, and string/float/nullary-after-erased shapes — **all correct**.

**One confirmed silent-wrong in the delta's scope, one pre-existing sibling with the same root, one minor over-conservative skip** — all filed as #1070 and being fixed on this branch before merge:

- The width recomputation classifies zero-size **type arguments** by the literal string `"Unit"` (`_eq_field_wasm_type`), so `Box<U>` (`type U = Unit;`), `Box<Future<Unit>>`, and alias chains silently read shifted offsets (0 where 22; nested variant 0 where 314) while the literal `Box<Unit>` is correct — declared fields dodge it via registration canonicalization, type arguments never canonicalize.
- Structural `Eq` over `Box<U>` compares equal structs unequal — pre-existing, same root; the #1060 round's "eq/show/hash were already correct" claim held only for the literal spelling and is corrected in the fix round.
- `_later_sub_pattern_reads` counts a zero-width `@Unit` binding as a later read (loud false-skip, correctness-preserving).

One erasure-aware fix at the shared classification closes the first two together; the fix round with RED pins from the review's exact shapes lands on this branch next.

aallan:
## Adversarial review record — delta pass over the #1070 commit

An independent adversarial review ran against 7d30ecf7 with instructions to hunt a seventh `== "Unit"` site, attack the erasure scoping in both directions (non-literal Unit spellings must work; non-Unit aliases must NOT be swept into the unit path), and stress the free-var/erasure classification boundary.

**Verdict on the commit: mergeable — all #1070-scoped surfaces correct.** The highest-value check passed: the new generator-side `_field_type_erases_to_unit` mirror and the canonical `_slot_name_erases_to_unit` agree on all 33 spellings fed from a shared alias table (refinement-over-Unit, cyclic aliases, `Future<CycA>`, chains, and every non-Unit alias) — the E613 gate and the `$eq` generator cannot desync in the #732 way. Wildcard reads, Eq (equal AND distinct), show, hash, and the trailing-binding rider all hold across `U`, `Future<Unit>`, `FU`, `U2=U=Unit`, `F=Future<U>`, `Future<Future<Unit>>`, nested-in-ADT, and mixed free-var/erased pairs; genuine free type vars still classify as free; non-Unit aliases keep exact offsets at values above 2^32. Conformance 160/160.

**The `== "Unit"` sweep found the six fixed sites plus two reachable misses — both pre-existing (verified on the parent commit), both now filed and folding into this PR:**

- **#1076 (silent-wrong).** `==` over a parameterized ADT whose type argument is a *non-Unit* alias (`Box<MyInt>`, `Box<MyStr>`, `Box<Future<Int>>`) silently pointer-compares — structurally equal values return 0 on a check-green program. `_type_arg_is_free_var` classifies any unrecognized single-segment name as a free variable; #1070 fixed the erases-to-Unit slice of that heuristic and this is its non-Unit half. Repro re-verified independently before filing.
- **#1077 (loud).** `show`/`hash` of `Tuple<U, Int>` and of a bare aliased-Unit value loud-skip: the Tuple-variadic plan branch passes raw `type_args` without `_canonical_unit_field_type`, and the top-level show/hash entry points compare the literal string `"Unit"` — the two call sites the six-site fix didn't reach. Degrades conservatively (drop, never a wrong value).

A third observation — element-wise `==` on arrays of parameterized ADTs falling back to pointer compare regardless of Unit/erasure — is being dupe-checked against the closed nested-generic-eq family before filing, since it is unrelated to this PR's mechanism.

Per the burndown's no-deferral rule the #1076/#1077 fixes ship on this branch before merge.

PR #1050, opened 2026-07-16
fix: infer element types for indexing builtin-call results (#1048)

### LINKED ISSUE #1048: Indexing a builtin-call result directly (array_concat(...)[i]) E602-drops the function for all element types
Indexing directly on a **built-in call result** is check-green but E602-drops the enclosing function — for every element type, not just `Future<T>`.

```vera
public fn g(-> @Int)
{ array_concat([1, 2], [3, 4])[2] }
```

`vera check` passes; `vera run --fn g` reports the function missing from exports (E602 skip). The let-bound control compiles and returns 3:

```vera
public fn f(-> @Int)
{
  let @Array<Int> = array_concat([1, 2], [3, 4]);
  @Array<Int>.0[2]
}
```

Root cause: the index-expression element-type inference (`_infer_index_element_type_expr`, vera/wasm/inference.py) resolves the collection's element type for user-function call results but has no arm for built-in call results, so the index read gets no WASM type and the function skips loudly.

Same check-green → E602 family as #1004/#1031/#1037/#1038/#1039; not Future-specific (plain `Int` elements reproduce). Found while validating the #1045 fix on PR #1041 (probe differential run against `main` at d084de0).

aallan:
## Summary

Closes #1048: indexing directly on a **built-in call result** (`array_concat([10,20],[30,40])[2]`) was check-green but E602-dropped the enclosing function for every element type — the index-expression element-type inference (`_infer_index_element_type_expr`) had a user-function-call arm but no built-in arm. The let-bound form always worked.

## The fix

The FnCall arm gains a built-in fallback (`_builtin_call_ret_named_type`) that reconstructs the call's return `NamedType` from the same shared consultor (`_get_arg_type_info_wasm` + the `_BUILTIN_PARAMETERIZED_RETURNS` table) the sibling Vera-type inference and monomorphizer discovery use — so the three cannot disagree on a builtin's return shape — and feeds it into the same `_alias_array_element` extraction as the user-function arm. Chained indexing on nested-array results (`array_concat([[1,2],[3,4]],[[5,6]])[2][0]`) works through the same round-trip.

## Coverage — stated precisely

Covers the **12** Array-returning builtins the consultor resolves: `array_concat`, `array_append`, `array_slice`, `array_filter`, `array_range`, `string_split`, `string_chars`, `string_lines`, `string_words`, `json_keys`, `md_extract_code_blocks`, `html_query`. The **8** whose element types are type variables (`array_map`, `array_mapi`, `array_flatten`, `array_reverse`, `array_sort_by`, `map_keys`, `map_values`, `set_to_array`) are absent from the consultor tables by design (instantiation-discovery lockstep) — their direct-index form still skips loudly and is tracked separately (see the follow-up issue linked from #1048); the let-bound form works for all of them. String-returning builtins are not index targets (the checker rejects `String` indexing).

## Validation

- **RED-first**: 6 tests (`TestIndexBuiltinCallResult`, `tests/test_codegen_arrays.py`) — four value tests across `array_concat`/`array_range`/`array_slice` and an `Array<String>` element (each RED as an E602 drop before the fix, exact element values), a structural no-E602 pin, and the let-bound control.
- **Mutation-validated**: reverting only the new arm flips exactly the five new tests RED with the control and the file's other 82 tests unaffected; restored, all green.
- **Full pre-commit gate green** on the first attempt (mypy, full pytest, conformance 160/160, examples 38/38, doc gates); counts oracle-reconciled (7,554).

Closes #1048


aallan:
## Adversarial review record

An independent reviewer attacked this PR together with the stacked #1051 fix (the two share the derivation machinery), hunting the class's nightmare case: a plausible-but-wrong element-type derivation (silent wrong value rather than a loud skip). ~42 probes across every mechanism class: comparator-return leakage into `array_sort_by` (element correctly from arg0), type-changing `array_map`/`mapi` closures (String→Int, Int→String, Int→Bool/Float64/Decimal/Option/user-ADT/refinement/`Array<Int>` with chained indexing — element always from the closure return), rep-swap discrimination on `Map<Int,String>` keys vs values, composition depth, generic-fn monomorphized contexts, boundary/OOB (traps, never garbage reads), and the pinned loud skips plus their siblings.

**Verdict: clean — no wrong derivation found; every shape is correct or degrades to the loud E602 skip.** Containment verified: the new derivations are reachable only from the index-element inference and never leak into the shared consultor or clone-name discovery (registry lock test unchanged).

Three loud-only pre-existing observations, all now tracked: alias-spelled derivation arguments (#1055 — index side fixed on the stacked #1051 branch; the call-emission side scoped into #1053), fn-typed slot closures into `array_map` (#1056, shared with the `apply_fn` slot path), and a user-fn-returning-Array argument asymmetry (scoped into #1053).

Issue #769, opened 2026-06-20
Monomorphizer type-inference / reindex completeness gaps (nested unify, builtin-return table, scope-aware De Bruijn)


aallan:
## Summary

Three pre-existing completeness gaps in the monomorphizer's type inference / substitution (`vera/monomorphize.py`, the logic relocated unchanged from `vera/codegen/monomorphize.py` by #732). All three make an *exotic* generic shape monomorphize to the wrong concrete type. They are **shared by codegen and the verifier** (the #732 verifier drives the same `Monomorphizer`), so they are not a #732 soundness regression — verify and codegen agree, both on the wrong instantiation, so `vera verify` clean still implies the runtime matches the (wrongly-monomorphized) behaviour — but the *program* does the wrong thing. Surfaced by the PR #767 review (CodeRabbit).

## 1. `_BUILTIN_PARAMETERIZED_RETURNS` is incomplete

`_get_arg_type_info` recognises only a small subset of builtins returning container types. Builtins such as `string_chars`, `string_split`, `string_lines`, `string_words`, `md_extract_code_blocks`, and the regex result helpers return parameterised types (see `vera/wasm/inference.py`) but are absent. A generic call like `first(string_chars("abc"))` cannot bind `T` from `Array<String>` and falls through to the `Bool` phantom default, producing the wrong monomorphisation.

## 2. Nested type-argument unification is shallow

`_unify_param_arg` binds a type variable only when the *immediate* type argument is the variable itself. For `@Array<Option<T>>` matched against `@Array<Option<Int>>`, `T` stays unbound (the matcher sees `Option<T>`, not `T`) and defaults to `Bool`. The same single-level matching appears in the FnType-alias inference. Walk the `TypeExpr` pairs recursively.

## 3. De Bruijn reindexing is parameter-only, not scope-aware

`_build_reindex_map` computes the collapse-reindex map from function *parameters* only, but Vera slot indices also count later `let` bindings, closure parameters, and match-pattern binders. If `A` and `B` both monomorphise to `Int`, a `let @B = ...` should shift later `@A.0` references to `@Int.1`; the parameter-only map leaves them `@Int.0`, silently changing semantics. The slot-name builder also collapses `Array<Option<A>>` and `Array<Option<B>>` to the same old name (it doesn't preserve nested type args). This one is the highest-severity of the three (a wrong slot read in a collapsed-type-var function with body bindings).

## Notes

[...]

PR #1061, opened 2026-07-16
fix: argument-derived element types — index derivations, call emission, container tags (#1051 #1052 #1053 #1055 #1063)

### LINKED ISSUE #1051: Direct indexing on type-variable-elemented builtin results (array_map, array_reverse, ...) still E602-drops
Split from #1048. The #1050 fix resolves direct indexing on builtin-call results for the 12 Array-returning builtins whose element types the shared consultor (`_get_arg_type_info_wasm` / `_BUILTIN_PARAMETERIZED_RETURNS`) can resolve. The remaining **8 Array-returning builtins have type-variable element shapes** and stay check-green but E602-drop when indexed directly:

`array_map`, `array_mapi`, `array_flatten`, `array_reverse`, `array_sort_by`, `map_keys`, `map_values`, `set_to_array`

```vera
public fn g(-> @Int)
{ array_reverse([10, 20, 30])[0] }
```

check-green; the function E602-drops (confirmed for `array_reverse` and `array_flatten`). The **let-bound form works** for all eight (the SlotRef arm resolves the declared type), so a one-line workaround exists.

Why not fixed in #1050: these builtins' element types are type variables — `array_map`'s element is the closure's return type, `map_keys`' is the map's key type — so resolving them at an index site requires argument-shape inference (closure-return / container-type-arg recovery), not a consultor-table row; and the consultor tables are kept in lockstep with monomorphizer instantiation discovery, so growing them affects discovery and clone naming. This is the same per-site type-recovery class as the #820 enabler work rather than a table gap.

Same check-green → E602 loud-skip family as #1048; split so the resolvable half didn't wait on the inference design.

### LINKED ISSUE #1052: Inline nested array literal as a call argument (array_flatten([[..],[..]])) E602-drops the function
An **inline nested array literal as a call argument** is check-green but E602-drops the enclosing function:

```vera
public fn f(-> @Int)
{
  let @Array<Int> = array_flatten([[10, 20], [30, 40]]);
  @Array<Int>.0[0]
}
```

`vera check` passes; the function is dropped (ArrayLit codegen skip: "could not recover inner element type"). The same call with a let-bound `@Array<Array<Int>>` argument compiles and runs.

Root cause: the array-literal element-type recovery has no arm for the nested-literal-as-call-argument position — the inner literals' element type is not recovered from the outer literal's context.

Same check-green → E602 loud-skip family as #1048/#1051. Found by the #1051 fix's validation (its flatten tests use let-bound nested arrays to sidestep this); reproduced on `main` at d084de0.

### LINKED ISSUE #1053: Type-variable builtin nested as a call argument (array_reverse(array_flatten(x))) E602-drops in call emission
A **type-variable-element builtin nested as another builtin's call argument** is check-green but E602-drops the enclosing function — even in fully let-bound form:

```vera
public fn f(-> @Int)
{
  let @Array<Array<Int>> = [[10, 20], [30, 40]];
  let @Array<Int> = array_reverse(array_flatten(@Array<Array<Int>>.0));
  @Array<Int>.0[0]
}
```

`vera check` passes; the function is dropped in the OUTER builtin's call emission ("could not infer array_reverse element type"). This is the call-emission sibling of #1051's index-element gap: the emission of `array_reverse(...)` needs its element type, and its argument is itself a type-variable-element builtin call the emission-side inference cannot type. #1051's fix (`_typevar_elem_builtin_ret_named_type`, PR chain #1050→) derives exactly these return types for the index path — the call-emission path needs the same derivation.

Same check-green → E602 loud-skip family as #1048/#1051. Found by the #1051 fix's validation; reproduced on `main` at d084de0.

### LINKED ISSUE #1055: Alias-spelled container args to the element-type derivations (array_flatten(@Grid.0)) E602-drop
An **alias-spelled container/array argument** to the #1048/#1051 element-type derivations E602-drops where the direct spelling works:

```vera
type Grid = Array<Array<Int>>;

public fn f(-> @Int)
{
  let @Grid = [[10, 20], [30, 40]];
  array_flatten(@Grid.0)[0]
}
```

check-green; the function drops. The direct `@Array<Array<Int>>.0` spelling compiles and runs. Root cause: the class-1/class-2 derivations (`_typevar_elem_builtin_ret_named_type` / `_named_type_from_arg_info`, vera/wasm/inference.py, PR chain #1050→#1051) inspect `.name`/`.type_args` on the argument's un-canonicalized alias NamedType — the same alias-blindness class as #1046/#1054, in the newest deciders. For `array_reverse` an aliased argument additionally drops earlier, in the reverse-call emission (the #1053 side).

Loud, never silent. Found by the adversarial review of the #1050+#1051 stack; index-side fix belongs on the #1051 branch, the call-emission side rides the #1053 work.

### LINKED ISSUE #1063: Container builtins over alias-spelled or user-fn args (map_values(@M.0)) emit with the wrong host tag: silent truncated values
**Silent wrong values**: `map_keys` / `map_values` / `map_get` / `set_to_array` over an **alias-spelled or user-fn-returned container** emit with the wrong host tag and return garbage — check- and verify-green, no diagnostics. Pre-existing on `main` (verified at d084de0):

```vera
type M = Map<String, Int>;

public fn f(-> @Int)
{
  let @M = map_insert(map_new(), "k", 8589934597);
  map_values(@M.0)[0]
}
```

returns **5** (the truncated low bytes) instead of 8589934597; an aliased `map_keys` over String keys returns 0 instead of the string; aliased `set_to_array` similarly truncates. Small payloads read correct-looking values by fresh-heap luck, which is how the shape evaded the existing pins.

Root cause: the container call emission derives its host-import tag from the argument's spelled type name; an alias (or user-fn call result) misses the container match and falls to the `"b"` (i32) tag — `map_values$vb` at the WAT level — so i64/pair payloads are read at the wrong width.

Same alias-blindness family as #1046/#1054/#1055/#1058/#1062, but silent-wrong rather than loud, and load-bearing for #1055: the aliased-container index pins on PR #1061 sit on top of this emission, so that PR must not merge without this fix (its small-value pins are luck-green).

Found by the #1052/#1053 fix round's argument-source sweep; fix built in the same round with big-value and String discriminating pins.

### LINKED ISSUE #1067: Array combinators halve alias-inner pair elements (Array<Row> where Row = Array<Int>): silent garbage
**Silent wrong values**: array combinators whose element type is itself an **alias to a pair type** halve every element:

```vera
type Row = Array<Int>;
type Grid = Array<Row>;

public fn f(-> @Int)
{
  let @Grid = [[1, 2], [3, 4]];
  array_reverse(@Grid.0)[0][0]
}
```

returns **4626322722586886145** (expected 3); `array_length(array_reverse(@Grid.0)[0])` returns 81948 (expected 2); the `array_flatten` depth-2 variant reads past its allocation (`[1]` length 0, `[1][0]` traps). All check- AND verify-green.

Root cause: the PR #1061 call-emission alias arm canonicalizes the COLLECTION name (`Grid` → `Array<Row>`) but hands the inner element name `"Row"` un-canonicalized to `_element_mem_size` (vera/wasm/helpers.py), whose unknown-name default is 4 bytes (ADT-pointer heuristic) — `Row` is an 8-byte pair, so the copy loop halves elements. The direct-spelling `@Array<Array<Row>>` twin mis-copies identically via a pre-existing AST walk on `main` (the mis-copy predates the branch; the branch's alias arm added new observation routes, so this is a merge blocker for #1061). `array_sort_by` with a `@Row`-typed comparator additionally traps loudly (indirect call type mismatch).

Fix shape: canonicalize the inner element name (the same `_canonicalize_alias_slot_name` walk) before the size/pair classification, at the emission consumers.

Found by the #1055-delta adversarial review of PR #1061 (probes p25/p26/p19/p19z/p28 in the review record).

### LINKED ISSUE #1068: Generic alias as declared return (MyMap<Int>) mis-derives container elements: invalid WASM, compile exits 0
**Invalid WASM emitted, compile exits 0**: a **generic alias as a declared fn return** feeding a container derivation produces a module that fails wasmtime validation — silently at compile time:

```vera
type MyMap<V> = Map<String, V>;

private fn mk(-> @MyMap<Int>)
{ map_insert(map_new(), "abcde", 99) }

public fn g(-> @Int)
{ string_length(map_keys(mk())[0]) }
```

`vera check` green, **`vera compile` exits 0**, and the first call traps `failed to compile: ... expected i32, found i64`. Root cause: `_get_arg_type_info_wasm`'s FnCall arm returns `("MyMap", ("Int",))` verbatim, and the PR #1061 class-2 container derivation takes `type_args[0]` as the key element **with no container-name check** — deriving `Int` (i64, stride 8) where the truth is `String`. `map_values(mk())` drops loudly instead — asymmetric. Introduced by the #1051 commit on that branch (proven by hunk-revert); merge blocker for #1061.

Fix shape: the class-2 derivation must verify the (canonicalized) container name IS `Map`/`Set` before consuming type args — a generic alias's own args are the alias's parameters, not the container's.

Secondary infrastructure gap worth its own look: `vera compile` hands over a module that fails wasmtime validation with no diagnostic — emission should validate (or surface the wasmtime error) rather than exit 0.

Found by the #1055-delta adversarial review of PR #1061 (probe p22).

### LINKED ISSUE #1071: Container builtins over ALIASED user-fn returns (fn mkm(-> @M)) silently truncate; Block-wrapped args hit the b-tag
**Silent wrong values (pre-existing on `main`)**: a user function whose declared return is an **alias of a container** silently truncates through the read-side container builtins:

```vera
type M = Map<String, Int>;

private fn mkm(-> @M)
{ map_insert(map_new(), "k", 8589934599) }

public fn g(-> @Int)
{
  let @Array<Int> = map_values(mkm());
  @Array<Int>.0[0]
}
```

check-green AND verify-green; returns **7** (= 8589934599 mod 2^32). Reproduces across map_keys / map_values / map_get / set_to_array (String keys read as length 0). The direct-spelled sibling (`-> @Map<String, Int>`) IS fixed by the PR #1061 commit; the aliased spelling is the exact intersection of the two argument sources that commit's prose claims closed.

Also in the same class: a **Block-wrapped container argument** (`map_values({ @M.0 })`) reaches the `"b"` tag with a non-empty collection — the argument-shape consultor has no Block/IfExpr arm.

Root cause (traced): `_get_arg_type_info_wasm`'s FnCall arm only accepts a declared return that is a NamedType **with** type args; a bare alias return (`NamedType("M")`) returns None before either the alias-canonicalizing branch or the user-fn extension can fire. Cheap fix shape: return `(ret_te.name, ())` for a bare-NamedType return so the existing alias branch resolves it. The `_map_wasm_tag` / `_container_arg_elem_name` comments over-claim ("before any arm can miss") and need softening or, better, making true.

Found by the tag-completeness adversarial review of PR #1061's container commit; merge-relevant for that PR (its comments claim closure of this exact shape).

aallan:
## Summary

Closes the **argument-derived element-type family** — five bugs across the index-inference and call-emission layers. Originally staged as two stacked PRs; folded into one when validation proved they interlock: the aliased-container index pins (#1055) sit on top of a container-emission mis-tag (#1063) and passed only by fresh-heap luck with small values — merging the index half alone would have turned a loud skip into silent wrong values for larger payloads.

## The five fixes

| Issue | Symptom (all check/verify-green) | Fix |
|---|---|---|
| #1051 | direct indexing on the eight type-variable-element Array builtins E602-drops | per-mechanism derivation (`_typevar_elem_builtin_ret_named_type`): argument-forwarding, container-args, closure-return — off the shared consultor (the #769 registry lock forbids TypeVar entries), reachable only from the index path |
| #1055 | alias-spelled arguments to the derivations E602-drop | the argument-type rebuilder canonicalizes bare alias names to full compound spellings (the #1037 walk) |
| #1052 | inline nested array literal as a call argument E602-drops | ArrayLit arm in `_translate_array_flatten`'s T-recovery |
| #1053 | type-variable builtins nested as call arguments E602-drop (plus aliased and user-fn argument spellings at call emission) | the combinator emission probes fall back to the shared #1051 derivation; alias canonicalization applied at the call-emission dual of #1055; user-fn nested returns read the registered declared return type (off the consultor — discovery untouched) |
| #1063 | **silent wrong values**: `map_keys`/`map_values`/`map_get`/`set_to_array` over aliased or user-fn container args fall to the `"b"` (i32) host tag — i64 values truncated, String keys garbled; pre-existing on `main` | the three container inference helpers consult the shared rebuilder first; values above 2^32 pin the tags; the genuinely-unknown empty-collection shape keeps its permissive fall-through, pinned |

## Validation

- **RED-first throughout**: 43 tests across the five issues — 12 direct-index value tests over all 8 builtins (type-changing `array_map` closures proven to derive from the closure **return**: Int→String variants discriminate where Int→Int would coincide), the #1052/#1053 shapes with exact values, the aliased/user-fn call-emission spellings, and the #1063 battery with **values above 2^32 and String keys** (the discriminating inputs whose absence let the mis-tag hide under small-value pins), plus structural no-E602 pins, chained indexing, loud-skip pins, and controls.
- **Per-site mutation matrices**: ten independent edit sites, each reverted alone flips exactly its own tests RED (including a sub-attribution proving the two #1053 sites non-redundant, and the closure-return coincidence-trap check).
- **Adversarial review**: the #1051 derivations were independently reviewed clean (~42 probes, record on PR #1050); the #1055 delta and the call-emission/container commits have a dedicated independent pass running, with its verdict to be appended to this thread before merge.
- **Full pre-commit gate** on the combined lineage (post-#1050 main); counts oracle-reconciled (7,622).

Closes #1051
Closes #1055
Closes #1052
Closes #1053
Closes #1063
Closes #1067
Closes #1068
Closes #1071


aallan:
## Adversarial review record

The #1051 derivations in this PR were independently adversarially reviewed as part of the #1050 stack (the review ran before this PR existed; full record on PR #1050): ~42 probes hunting plausible-but-wrong element-type derivations across all three mechanism classes — comparator-return leakage, type-changing closures, container rep-swaps, composition depth, monomorphized contexts, OOB boundaries, loud-skip siblings. **Clean: every shape correct or loudly skipped; containment into the index path verified against the discovery registry lock.**

The #1055 commit (alias-spelled argument canonicalization) postdates that review, so a focused independent pass over that delta — alias chains, parameterized aliases, aliased containers across all three classes, and the loud-skip boundary claims — is running now; its outcome will be recorded here.

aallan:
## Adversarial review record — #1055 delta verdict (and branch findings)

The dedicated pass over the #1055 alias-canonicalization delta is complete. **The delta itself is clean**: ~30 probes (alias chains to 3 hops, rep-swap discrimination on `Map<Int,String>` keys-vs-values with direct-spelling twins, two rep-opposed aliases in one program, value-spelling aliases across scalar/String/Float64, negative-space rejections, eager-GC re-runs) — every canonicalization produced the correct compound spelling; the block is mutation-proven load-bearing; no cross-talk.

The sweep ran at the branch tip (which had gained the call-emission/container commit mid-review) and confirmed **three findings elsewhere on the branch**, now filed and being fixed here before merge:

- **#1067 (merge blocker, silent-wrong)**: combinators halve **alias-inner** pair elements (`Array<Row>`, `Row = Array<Int>` — reverse/flatten return garbage or read past allocations). The collection name canonicalizes but the inner element name reaches the size classification raw; the direct-spelling mis-copy predates the branch on `main`, but this branch's alias arm adds new routes to it.
- **#1068 (merge blocker, invalid-WASM)**: a generic alias as a declared return (`MyMap<Int>`) feeds the class-2 container derivation, which consumes its type args with no container-name check — `vera compile` exits 0 on a module that fails wasmtime validation. Also surfaces a standalone infrastructure gap (compile should not hand over a validation-failing module silently).
- **#1069 (minor)**: an E121 message leaks a raw type variable.

Depth-1 aliased shapes (the PR's pins) are all correct; the pins' small-value luck-green issue found earlier is already fixed by the #1063 commit with >2^32 discriminators. The fix round for #1067/#1068 lands on this branch next; the parallel tag-completeness review of the container commit is still running and reports here when done.

aallan:
## Adversarial review record — call-emission/container commit verdict

The dedicated pass over the third commit (ef1f81a4, #1052/#1053/#1063) is in: **no regression introduced; the array-literal recovery and the fixed container shapes are sound.** The #1052/#1053 work held across mixed inner lengths, three-level nesting, String and big-i64 elements, both nesting directions, and the empty-first-inner boundary (loud). The six element-expression-tagged builtins (map_insert/contains/remove, set_add/contains/remove) are correctly unaffected by container spellings. All pins hold under eager GC. The direct-spelled user-fn container return is confirmed **fixed by this commit** (7 → 8589934599).

Two pre-existing findings at the fix's own boundary, filed as #1071 and being fixed on this branch: an **aliased user-fn return** (`fn mkm(-> @M)`) still silently truncates through the read-side container builtins (the FnCall consultor arm requires type args, so a bare alias return exits before the alias branch fires), and a **Block-wrapped container argument** reaches the `"b"` tag non-empty. The commit's `_map_wasm_tag`/`_container_arg_elem_name` comments over-claim closure ("before any arm can miss") — the fix round makes them true.

With this, all three commits on the PR have independent adversarial verdicts; the consolidated #1067/#1068/#1071 fix round lands next, then this PR is merge-ready.

aallan:
## Adversarial review record — delta pass over the three-bug fold commit

An independent adversarial review ran against 4085042c (the #1067/#1068/#1071 commit), with explicit instructions to attack the chained-generic-alias seam the commit's own author flagged as residue, plus large-value coincidence traps and alias-vs-direct differentials.

**Verdict: mergeable — no silent-wrong found.** Highlights of the probe battery (~30 shapes, all values above 2^32):

- **The chained-generic-alias seam is NOT silent-wrong reachable.** The report's exact worry (`type A<V> = B<V>; type B<V> = Map<String, V>`) resolves correctly through map_values/map_keys, including three-level chains, parameter renaming, partial application (`Outer<V> = Inner<String, V>`), Set chains, and a discarded-parameter outer. The free-var leak in `_canonical_named_type`'s non-substituting follow is only reachable when an alias is applied at the wrong arity — and every such spelling is already E133+E121 at check time; even with the checker bypassed the leaked tag produces an invalid module (loud), never a wrong value, because the construction side always tags concretely and WASM validation rejects the disagreement.
- #1067's element canonicalization holds for alias-of-alias chains and direct-spelling twins across reverse/concat/flatten/sort_by; #1071's Block tail-unwrap holds for nested blocks, lets before the tail, tail-is-a-call, and shadowed inner lets; the #1068 container-name verification keeps unresolvable shapes as loud skips.
- 746 tests across the touched and adjacent suites pass; aliased and direct spellings agree in every differential.

**One pre-existing observation, now filed with a vehicle: #1075.** A Map/Set with a zero-size value type (`Map<String, Unit>`) checks clean but emits an invalid WASM module — reproduced identically on main and on this branch, for both the direct and alias spellings (the branch's alias resolution correctly makes them behave the same). It originates in zero-size value emission, not this commit's alias machinery; per the burndown's no-deferral rule the fix folds into this PR next.

Methodology note for the record: the review initially invalidated ~5 probe rounds after discovering that script-mode probe runs silently imported the main checkout's compiler via the editable install; every reported result above was re-run with the worktree pinned and asserted on `vera.__file__`.

Issue #1076, opened 2026-07-16
Structural == over non-Unit alias type arguments (Box<MyInt>) silently pointer-compares: equal values compare unequal


aallan:
## Summary

Structural `==` over a parameterized ADT whose type argument is spelled as any **non-Unit alias or transparent wrapper** — `Box<MyInt>` (`type MyInt = Int;`), `Box<MyStr>`, `Box<MyBool>`, `Box<Future<Int>>` — silently falls back to a scalar **pointer compare**: two structurally equal values compare unequal on a check-and-verify-green program. **Silent wrong value.**

## Repro

```vera
type MyInt = Int;

private data Box<T> { MkB(Int, T, Int) }

private fn a(-> @Box<MyInt>)
  requires(true) ensures(true) effects(pure)
{ MkB(11, 7, 22) }

public fn f(-> @Bool)
  requires(true) ensures(true) effects(pure)
{
  let @Box<MyInt> = a();
  let @Box<MyInt> = a();
  @Box<MyInt>.0 == @Box<MyInt>.1
}
```

- `vera check` → clean.
- `vera run --fn f` → **0** (unequal). Expected **1**: the two values are structurally identical. The literal spelling `Box<Int>` of the same program returns 1.

## Root cause

`_type_arg_is_free_var` (vera/wasm/operators.py:473) returns `True` for any single-segment type-argument name that is neither a concrete base, a registered ADT, nor erases-to-Unit — so a non-Unit alias is classified as a free type variable and `==` takes the scalar pointer-compare fallback instead of structural field comparison. This is the **non-Unit half** of the heuristic #1070 fixed: #1070 keyed the erases-to-Unit spellings on erasure and left every other alias spelling mis-classified.

## Pre-existing

Verified identical on the parent commit (7e1c40d2, before the #1070 fix) — #1070 neither introduced nor worsened it. Not previously tracked. Found by the adversarial delta review of the #1070 commit on PR #1049; the fix ships on the same branch.

aallan:
Fix in review: PR #1084 (targets release/v0.1.5 — GitHub does not render the close link for non-default-branch PRs; the PR body carries Closes #1076 and this issue closes when the release train lands on main).

PR #1080, opened 2026-07-16
fix: concrete scrutinee instantiations for the type-param wildcard walks (#1065, #1072, #1073)

### LINKED ISSUE #1065: Direct-call match scrutinee with a harmful generic wildcard loud-skips: thread expr_semantic_types to compile it
A **direct-call match scrutinee** (`match mk() { ... }`) whose type-parameter wildcard is followed by a later field read loud-skips (E602) rather than compiling: the codegen-side `_infer_vera_type(FnCall)` returns the base name without type arguments (`"Box"`, not `"Box<Unit>"`), so the #1060 instantiation-aware wildcard walks cannot recover the erased field's width and correctly refuse to guess (a wrong offset would corrupt the later read — the pre-#1060 behavior).

The SlotRef/let-bound scrutinee form recovers correctly; a trailing unrecoverable wildcard (width never consumed) still compiles. No test/conformance/example corpus case hits the skip.

Completeness direction: thread the checker's `expr_semantic_types` (which carries the full `Box<Unit>`) into the match lowering, turning the loud skip into a correct compile. Deferred from PR #1049's #1060 round as higher-risk than the fix warranted; both behaviors are pinned there (`test_direct_call_unrecoverable_read_after_loud_skips`, `..._trailing_wildcard_compiles`).

### LINKED ISSUE #1072: Generic-call match scrutinee: wildcard walks cannot recover the instantiation (silent wrong value on main)
## Summary

A **generic-fn-call `match` scrutinee** whose declared return is a parameterized type cannot supply the concrete instantiation to the type-parameter wildcard walks. On `main` this is a **silent wrong value** (the #1060 bug class through the generic-call door); on the #1049 stack (#1043 + #1060, plus the #1065 non-generic direct-call fix) it is a **sound E602 loud-skip** — never a wrong value, but a check-green program that should compile and run correctly does not.

Found during the #1065 fix: the `_match_scrutinee_vera_type` helper added there gates on **non-generic** calls (`scrutinee.name not in _generic_fn_info`), because a generic callee's declared return still carries unresolved type variables. This issue is the generic sibling: resolve those variables from the call site.

## Reproduction

```vera
private data P2<A, B> { MkP2(Int, B, Int) }

private forall<T> fn wrap(@T -> @P2<T, Unit>)
{ MkP2(11, (), 22) }

public fn main(-> @Int)
{
  match wrap(5) {
    MkP2(@Int, _, @Int) -> @Int.0
  }
}
```

`vera check` passes (T resolves to `Int`; the erased field is the concrete `Unit` type argument in the declared return).

**Expected:** compiles; `vera run` prints `22` (the wildcard over the erased `B = Unit` field advances 0 bytes; `@Int.0` is the trailing 22).

**Actual on `main` (796a016e):** `vera run` prints `0` — silently wrong. The wildcard walk advances by the generic 4-byte i32 placeholder, so the trailing `Int` is read 4 bytes high on a check-green program with no diagnostics.

**Actual on the #1049 stack + #1065:** E602 loud-skip (sound, but the shape family should compile):

```
Function 'main' body contains unsupported WildcardPattern: wildcard over
type-parameter field 1 of 'MkP2': concrete instantiation unrecoverable from
scrutinee type 'P2', and a later field is read at an offset that depends on
this field's erased width — function skipped.
```

## Validated sibling shapes (all check-green, all E602 on the stack)

- `forall<T> fn wrap(@T -> @Box<T>)` called as `wrap("x")` — the wildcarded field IS the forall var, at `String` (i32_pair width, discriminating against the generic 4-byte placeholder).
- `forall<T> fn wrap(@T -> @Box<Unit>)` — a fully **concrete** parameterized return on a generic fn: no variables to resolve at all, still skipped by the non-generic gate.
- The nested-constructor and `String`-read-back variants of the primary repro (`En(_, MkBox(@Int))` over `EntryAB<Int, Unit>`; `N(_, @String)` over `NamedAB<Int, Unit>`).

## Shapes that are NOT in scope (checker-rejected, so never check-green)

- Instantiating the forall var itself at `Unit` (`wrap(())` against `@T -> @Box<T>`) is **E206**-rejected — the generic Unit-erasure shape arrives only via a concrete `Unit` type argument in the declared return, as above.
- A phantom-var callee (`forall<T> fn make(-> @Box<T>)` called as `make()`) is **E121**-rejected, so no check-green genuinely-unresolvable shape remains in the family.

## Root cause

`_infer_vera_type(FnCall)` routes a generic call through `_infer_fncall_vera_type`, whose generic arm substitutes only a **bare type-var** return (`sub_map.get(decl_ret.name, decl_ret.name)`): for a parameterized return like `@P2<T, Unit>` it returns the bare base head (`"P2"`), dropping the type arguments the #1060 instantiation-aware walk needs to recompute the erased field width (same bare-head collapse as the #911 branch for non-generic parameterized returns, which the #1065 fix bypassed for non-generic callees only). `_match_scrutinee_vera_type` (vera/wasm/data.py) therefore falls through to the bare head for a generic callee, and `_wildcard_field_wasm_type` correctly refuses to guess.

## Fix direction

Resolve the callee's declared-return type variables from the call site — the same `_unify_param_arg_wasm` unification `_infer_fncall_vera_type` already performs for generic calls — substitute them into the declared return TypeExpr, and render the full name (`P2<Int, Unit>`) exactly as the non-generic arm does via `_format_named_type`. Any unresolved variable falls back to the current loud-skip (sound).

Fix ships on the #1049-stacked #1065 branch (`fix/1065-semantic-type-threading`), same subsystem, per the burndown rule.

### LINKED ISSUE #1073: Module-call match scrutinee: wildcard walks cannot recover the instantiation (silent wrong value on main)
## Summary

The **module-call door** of #1072: a `match` scrutinee that is a **module-qualified call** (`boxlib::mk()`) cannot supply its concrete instantiation to the type-parameter wildcard walks. On `main` this is a **silent wrong value**; on the #1049 stack (+ the #1065 non-generic and #1072 generic FnCall fixes) it is a **sound E602 loud-skip** on a check-green program.

Found while validating #1072: `_match_scrutinee_vera_type` (vera/wasm/data.py) dispatches on `ast.FnCall`, so an `ast.ModuleCall` scrutinee never enters either recovery arm and falls through to `_infer_vera_type`, whose ModuleCall arm delegates to `_infer_fncall_vera_type` — the same bare-base-head collapse (`"Box"`, type args dropped) as the other doors.

## Reproduction

`boxlib.vera`:

```vera
module boxlib;

public data Box<T> { MkB(Int, T, Int) }

public fn mk(-> @Box<Unit>)
{ MkB(11, (), 22) }
```

`modmain.vera` (same directory):

```vera
import boxlib;

public fn main(-> @Int)
{
  match boxlib::mk() {
    MkB(@Int, _, @Int) -> @Int.0
  }
}
```

`vera check modmain.vera` passes.

**Expected:** `vera run modmain.vera` prints `22`.

**Actual on `main` (796a016e):** prints `0` — the wildcard over the erased `Unit` field advances by the generic 4-byte placeholder and the trailing `Int` is read 4 bytes high, silently.

**Actual on the #1049 stack + #1065/#1072:** E602 loud-skip (sound, never a wrong value):

```
Function 'main' body contains unsupported WildcardPattern: wildcard over
type-parameter field 1 of 'MkB': concrete instantiation unrecoverable from
scrutinee type 'Box', and a later field is read at an offset that depends on
this field's erased width — function skipped.
```

## Fix direction

Mirror `_infer_vera_type`'s ModuleCall arm inside `_match_scrutinee_vera_type`: resolve the qualified target through the single shared resolver (`_resolve_module_call_wasm_name`, the #774-reviewed source of truth that consumes `path`) and recurse with the resolved-name `FnCall`. The resolved name lands in the registries the existing arms already consult — the bare name or `mod$…` alias in `_fn_ret_type_exprs` for a non-generic import, a per-instantiation clone name (whose registered declared return is already substituted) for a shadowed generic — and anything unregistered falls through to the current sound loud-skip.

Fix ships on the #1049-stacked #1065 branch (`fix/1065-semantic-type-threading`) alongside #1072, same subsystem, per the burndown rule.

aallan:
Threads concrete scrutinee instantiations into the #1060 type-param wildcard walks, closing three check-green E602 loud-skips (one of which is silent-wrong on released v0.1.4):

- **#1065 — direct-call scrutinee.** `match mk() { ... }` where the callee's declared return is a parameterized type erasing a field to Unit (`Box<Unit>`) loud-skipped the enclosing function. The fix renders the callee's declared return `TypeExpr` in full (byte-identical to the SlotRef path), so the #1060 walks see the concrete instantiation. Non-generic direct calls only; the resolver falls through to the sound loud-skip when it cannot prove the type.
- **#1072 — generic-call scrutinee.** `match wrap(5) { ... }` (declared return still carrying type variables) unifies the callee's params against the call arguments with the same `_unify_param_arg_wasm` the generic call-rewrite uses, then renders the substituted return; an unbound variable renders nothing and keeps the loud skip. On released v0.1.4 this shape ran and silently returned 0.
- **#1073 — module-call scrutinee.** `match boxlib::mk() { ... }` resolves through the shared module-call resolver and recurses with the resolved-name call. Same silent-0 history on v0.1.4, found while validating #1072.

Validation: RED→GREEN per shape (trailing-Int 22, nested-ctor 314, String read-back, module 22, imported-generic compound 22 — all verified end-to-end through `vera run`); per-arm mutation matrices (generic arm off → exactly 6 tests RED incl. the imported-generic compound; module arm off → exactly the 2 module tests RED); full pre-commit gate green on both commits.

**Stacked on #1049** (this branch builds on the #1060 instantiation-aware walks). Retarget to `release/v0.1.5` and rebase when #1049 merges.

Closes #1065
Closes #1072
Closes #1073


aallan:
## Adversarial review record

An independent adversarial review ran against the branch (7af8d8d0 + 36891345, based on the pre-#1070 tip of #1049), attacking the unification with swapped-param/nested/two-arg shapes, unbound-var soundness, module-arm recursion, and the recovered-instantiation width path.

**What holds:** unification renders the correct instantiation on every order-attack (swapped params, param-used-twice, arg-is-itself-a-generic-call, nested Unit-in-type-arg); phantom or unbindable return vars are E121-rejected or degrade to a sound loud-skip — never a guessed type; the module arm recurses exactly one hop with the imported-generic compound correct; non-erased widths (Bool/Float64/String/ADT) match construction on the newly enabled paths; the 201-test battery passes.

**One confirmed finding: for a recovered type argument spelled as an alias, alias chain, or `Future<Unit>`, the fix converts the base's sound E602 loud-skip into a silent wrong read** (`match wrap(mkAge()) { MkW(_, @Int) -> ... }` with `type Age = Int` reads 5 for 22; `Box<U>` and `Box<Future<Unit>>` direct-call shapes read 0). Root cause is not this branch's new code: the wildcard-width recomputation feeds the raw recovered spelling to `_eq_field_wasm_type`, whose literal-`Unit` test and i32 fallback predate the PR — the erasure half of that defect is fixed by the merged #1070 (in `release/v0.1.5`, absent from this branch's base), and the non-Unit-alias half by #1076 (in PR #1084). The three verified repros are preserved and become this branch's regression tests.

**Disposition:** merge #1084 first; rebase this branch onto the resulting `release/v0.1.5`; re-run the three repros against the rebased tree (expected: correct values via the composed #1070+#1076 machinery — verified empirically, not assumed); pin them as tests; re-gate. Not merge-ready until that verification passes.

aallan:
Review disposition resolved: the branch is squashed and rebased onto release/v0.1.5 (post-#1049/#1084), and the review's three regression repros — recovered type arguments spelled as an Int alias, a Unit alias, and Future<Unit> — all return the correct values on the rebased tree (verified by running the review's own probe programs, then pinned as three regression tests guarding the #1070+#1076 composition). One #1070-era boundary pin flipped accordingly: the direct-call read-after-erased-binding shape was pinned as a loud skip while unrecoverable and now resolves via the declared-return recovery, so it asserts the computed value; the loud-skip boundary stays pinned by the unresolved-variable shapes. Full gate and the complete wall are green at d2b0b3ce.

PR #1084, opened 2026-07-16
fix: ground alias type arguments and index-expression operands in the Eq dispatch (#1076, #1077, #1078)

### LINKED ISSUE #1076: Structural == over non-Unit alias type arguments (Box<MyInt>) silently pointer-compares: equal values compare unequal

### LINKED ISSUE #1077: show/hash of Tuple<U, Int> (aliased-Unit type arg) and of bare aliased-Unit values loud-skip the function
## Summary

`show` / `hash` over a **tuple whose type argument is a non-literal erases-to-Unit spelling** (`Tuple<U, Int>` where `type U = Unit;`), and over a **bare value of an erases-to-Unit alias type**, loud-skip the enclosing function (E602-class drop) — while the literal spellings (`Tuple<Unit, Int>`, bare `Unit`) render fine. Check-green, then the function is dropped.

## Repro

```vera
type U = Unit;

private fn mkt(-> @Tuple<U, Int>)
{ Tuple((), 42) }

public fn f(-> @Int)
{
  let @Tuple<U, Int> = mkt();
  string_length(show(@Tuple<U, Int>.0))
}
```

- `vera check` → clean.
- `vera run --fn f` → "No exported functions to call" (the function was loud-skipped). With `Tuple<Unit, Int>` the same program renders `(unit, 42)` and returns its length.

## Root cause

Two sites, same missing canonicalization:

- `_composite_ctor_plans`'s Tuple-variadic branch (vera/wasm/calls_handlers.py:522-530) passes **raw `type_args`** as the field type without routing through `_canonical_unit_field_type` — unlike the registered-ADT branch (line 556), which #1070 fixed.
- The top-level `_translate_show` / `_translate_hash` (vera/wasm/calls_handlers.py:72 / 143) compare the inferred type name against the literal string `"Unit"`, so a bare `U`-typed value misses the Unit arm.

Both degrade conservatively (loud drop, never a wrong value — the tuple *layout* itself is correct, verified by a discriminating trailing-field read). The Tuple case is squarely in #1070's stated domain (an erases-to-Unit type argument spelled non-literally) — these are the two show/hash call sites the six-site fix didn't reach.

## Pre-existing

Loud-skips identically on the parent commit (7e1c40d2). Found by the adversarial delta review of the #1070 commit on PR #1049; the fix ships on the same branch.

### LINKED ISSUE #1078: Element-wise == on arrays of parameterized ADTs (Array<Box<Int>>) silently pointer-compares: equal elements compare unequal
## Summary

Element-wise `==` on an array of **parameterized** ADT values — `Array<Box<Int>>`, literal spelling included — silently falls back to a scalar **pointer compare**: two structurally equal elements compare unequal on a check-green program. **Silent wrong value.** Arrays of **non-generic** ADTs (`Array<P>`) compare structurally and correctly, as does direct (non-array) `==` on the same parameterized values.

## Repro

```vera
data Box<T> { MkB(Int, T, Int) }

private fn mkbox(-> @Box<Int>)
{ MkB(11, 7, 22) }

public fn f(-> @Bool)
{
  let @Array<Box<Int>> = [mkbox(), mkbox()];
  @Array<Box<Int>>.0[0] == @Array<Box<Int>>.0[1]
}
```

- `vera check` → clean.
- `vera run --fn f` → **0** (unequal). Expected **1**: the two elements are structurally identical. The same values compared directly (two `Box<Int>` slots, no array) return 1, and an array of a non-generic ADT (`data P { MkP(Int, Int) }`, `Array<P>`) returns 1.

Same wrong result for `Array<Box<Unit>>` and `Array<Box<U>>` (`type U = Unit;`) element comparisons — the spelling does not matter; this is an argument-**drop**, not an alias-resolution issue.

## Root cause

`_infer_vera_type` on an `IndexExpr` operand returns the element type's **bare head** — `_infer_index_element_type` (vera/wasm/inference.py:1373) computes the full element `NamedType` via `_infer_index_element_type_expr` and then returns `te.name`, dropping `type_args` (`Box<Int>` → `"Box"`). The `==` dispatch's recovery chain `_eq_operand_full_name` (vera/wasm/operators.py:650) has arms for `ConstructorCall` operands, bare generic-ADT slots, and truncated clone names — but no `IndexExpr` arm — so the bare `"Box"` reaches `_is_lost_type_arg_clone`, is classified as a lost-type-argument clone (the #772/#912 residue class), and the comparison takes the scalar pointer-compare fallback.

The full element type is recoverable at the site: the indexed collection (`@Array<Box<Int>>.0`) carries its complete type arguments, and `_infer_index_element_type_expr` already returns the full `NamedType` — only the string conversion truncates it.

## Pre-existing

Verified identical on 7e1c40d2 (parent of the #1070 fix): `vera run --fn f` → 0. Unrelated to the Unit/erasure family (#1043/#1060/#1070) and not covered by the closed nested-generic-eq family (#921 compare/Ord pointer-compare; #923/#932 loud E613 divergences; #934 generic-mutual-recursion) — none of those describes the array-element operand path. Found by the adversarial delta review of the #1070 commit on PR #1049; the fix ships on the same branch.

aallan:
Three bugs in the structural-Eq dispatch, all check-green with wrong or missing behavior:

- **#1076 — silent-wrong.** `==` over a parameterized ADT whose type argument is any *non-Unit* alias or transparent wrapper (`Box<MyInt>`, `Box<MyStr>`, `Box<Future<Int>>`) fell back to a scalar **pointer compare**: structurally equal values compared unequal. The free-var classifier and concreteness gate treated any unrecognized single-segment name as an unresolved type variable — the non-Unit half of the heuristic #1070 fixed. All classification sites now ground a type argument through canonicalization before deciding (a genuine free `T` still classifies free, pinned).
- **#1077 — loud.** `show`/`hash` of `Tuple<U, Int>` (aliased-Unit type argument) and of bare aliased-Unit values E602-dropped: the Tuple-variadic plan branch passed raw type args without the erasure canonicalization, and the top-level entries compared the literal string `Unit`. Both routes now canonicalize; the shapes render like their literal spellings.
- **#1078 — silent-wrong.** Element-wise `==` on arrays of parameterized ADTs (`Array<Box<Int>>`, literal spelling included) pointer-compared: the operand recovery chain had no IndexExpr arm, so the element's bare `Box` head classified as a lost-type-argument clone. A new arm re-derives the full element spelling from the indexed collection (the full NamedType was already computed; only the string conversion truncated it).

Validation: 34 new tests plus wildcard-width and control pins (equal AND distinct pairs, values above 2^32, low-bits-collision distinct cases, content-compared String fields, non-generic-array and direct-compare controls); an eight-mutation matrix with exact per-site RED sets; the eq gate↔codegen differential; full suite and gate green (7,800 tests, 126 files).

**Stacked on #1049.** Retarget to `release/v0.1.5` when it merges.

Closes #1076
Closes #1077
Closes #1078


aallan:
## Adversarial review record

An independent adversarial review ran against e9ee93f2, attacking over-grounding of genuine free type variables, the E613↔eq-generator lockstep, IndexExpr recovery depth, width honesty at colliding low bits, and show/hash rendering blast radius.

**Verdict: solid — could not break it.** Every PR result matched ground truth: genuine free-T shapes still lower to the harmless scalar path; the two mirror grounders agree across alias chains, Future-of-alias, cycle-cut self-references, and nested compounds (hand-traced and differentially confirmed); the IndexExpr arm never mis-derives (stays loud on FnCall/map_values/nested collections it cannot prove); all distinct-pair pins hold at 2^32-colliding payloads; non-Unit tuple rendering is byte-identical to base; 1,488 tests and all 160 conformance programs pass. The PR also closes base-side holes: the base silently pointer-compared several non-Eq aliased types that now correctly E613.

**Three pre-existing defects surfaced (byte-verified identical on the base — not this PR's changes), filed with the reviewer's verified repros:** #1085 (direct `==` over an alias of a whole generic ADT pointer-compares — the un-grounded whole-operand sibling of this PR's type-argument fix), #1086 (`forall<T where Eq<T>>` at an Eq alias wrongly E613s — the constraint gate doesn't ground), #1087 (`show`/`hash` of bare `Future<Unit>`/aliased-Future values loud-skip before the erasure-keyed arm). All three join the successor queue.

Issue #1074, opened 2026-07-16
Array elements spelled Future<Alias> are mis-sized: silent wrong Bool/Byte values, invalid-WASM traps for scalars


aallan:
## Summary

An array element spelled `Future<A>` where `A` is a **type alias** — written directly (`Array<Future<FlagA>>`) or hidden behind another alias (`type X = Future<A>; ... Array<X>`) — is mis-sized as a 4-byte i32 element. `check` and `verify` both pass.

- **Bool/Byte payloads → silent wrong values**: the 1-byte-packed source is read at a 4-byte stride, so every element misreads (e.g. all elements come back `true`).
- **Int/Nat/Float64/String payloads → WASM-validation trap** (`expected i32, found i64` and variants) on a program the compiler reports as successfully compiled — invalid WASM behind exit 0.

## Repro (silent-wrong, Bool payload)

```vera
type FlagA = Bool;

public fn f(-> @Int)
  requires(true) ensures(true) effects(<Async>)
{
  let @Future<Bool> = async(false);
  let @Future<Bool> = async(true);
  let @Array<Future<Bool>> = [@Future<Bool>.1, @Future<Bool>.0];
  let @Array<Future<FlagA>> = @Array<Future<Bool>>.0;
  let @Array<Future<FlagA>> = array_concat(@Array<Future<FlagA>>.0, @Array<Future<FlagA>>.0);
  ...
}
```

Reading the four concatenated elements should yield the digit pattern `2121`; the program returns `1111` (every element misread as `true`). `array_append` shows the same shape (`211` expected, `111` actual). Swapping the payload to an Int alias with a value above 2^32 turns the same spelling into a WASM-validation trap at instantiation.

## Root cause

The element-type deciders are name-keyed, and the alias canonicalization added on the `fix/1039-pair-future-lets` branch resolves only a **top-level** alias name (or peels the `Future<>` wrapper) — it never canonicalizes an alias sitting **inside** a compound spelling's type arguments. Instrumented directly: `_infer_concat_elem_type` returns the string `Future<FlagA>`, and `_element_mem_size("Future<FlagA>")` = 4 with `i32.store`, where the correct canonical `Future<Bool>` gives 1 with `i32.store8`. `_strip_future` (vera/wasm/helpers.py:251) peels the wrapper and hands the bare, unresolved alias to the size dict, which defaults to 4.

Affected sites (branch `fix/1039-pair-future-lets` at cf410e3d):

[...]

PR #1089, opened 2026-07-16
fix: render full Future spellings in map/mapi/fold closure-return inference (#1079, #1081, #1082)

### LINKED ISSUE #1079: array_map/mapi/fold over Array<Future<T>> mis-size the output element: bare-head Future in closure-return inference
## Summary

`array_map` / `array_mapi` / `array_fold` over an `Array<Future<T>>` mis-size the output element (or fold accumulator): the closure-return inference returns the **bare head `"Future"` with the type argument dropped entirely**, so the element-size deciders fall to the 4-byte i32 default while an i64 or pair payload occupies 8+ bytes. Check and verify pass; the compiled module fails WASM validation at run (`type mismatch`) for Int/Nat/Float64/String payloads — a check-green program that reports successful compilation and then traps. A 1-byte Bool payload likely misreads silently instead (same stride family as #1074's Bool case).

## Repro

A pure identity closure forwarding a `Future<Int>` input, mapped over a two-element array whose first payload exceeds 2^32:

```vera
(see the probe: an fn(@Future<Int> -> @Future<Int>) effects(pure) { @Future<Int>.0 }
identity mapped over [async(5000000001), async(22)], then the mapped elements awaited)
```

- `vera check` → clean (`ok=true`).
- `vera run` → WASM validation trap: `type mismatch`.

Note the reachability constraint: `array_map` rejects an *async* closure, so the trigger is forwarding an existing `Future` value through a **pure** closure — the identity/reshuffle pattern.

## Root cause

`_infer_closure_return_vera_type` (and `_infer_fold_init_vera_type`) return `canonical.name` — the bare `"Future"` — discarding the payload spelling, so `_element_mem_size("Future")` = 4 and `_element_wasm_type` = `i32`. Instrumentation on the `fix/1039-pair-future-lets` branch confirmed the inferred name and the 4/i32 sizing.

## Distinct from the branch's fixed bugs

This is the **bare-head family (#1057's mechanism) at a different site** — the map/mapi/fold closure-return and accumulator inference — which the #1045/#1057 element fixes and the #1074 payload canonicalization never reach (this path returns `.name` before the deciders run). It mis-sizes the **direct** `Future<Int>` spelling too, not only aliased payloads. Found during the #1074 fix on PR #1041; verified check-green-then-trap on that branch at b9c897e1.

### LINKED ISSUE #1081: array_map over Array<Future<Bool>> silently misreads every element: read/write stride desync (regression on main, post-#1041)
## Summary

`array_map` over an `Array<Future<Bool>>` — direct spelling, no aliases, a pure identity closure — **silently misreads every element** on current `main`: a `[false, true, false, true]` array maps to all-false. Check and verify green, exit 0. This is a **regression introduced by the #1041 merge**: the same program returned correct values on the pre-merge base.

Released v0.1.4 is NOT affected (the regression post-dates the tag). It is live on `main` only.

## Repro (verified as a base-vs-post differential)

A pure identity closure `fn(@Future<Bool> -> @Future<Bool>) effects(pure) { @Future<Bool>.0 }` mapped over a four-element `[F, T, F, T]` array of already-created futures, then a digit read-back of the four elements:

- Pre-#1041 base (cd2da2bb): **2121** (correct).
- Current main (fa3c79ab): **2222** (every element false).
- Control (same array, no map): 2121 on both — the regression is isolated to the `array_map` step.

Probe files: `map_identity_4.vera` and `ctrl_nomap_4.vera` (adversarial-review scratchpad, `advF-1074/`).

## Root cause

A read/write stride desync. #1041's element-READ path fix (`_strip_future`, the #1045/#1057 family) now correctly sizes a `Future<Bool>` element read at 1 byte — but `array_map`'s element-WRITE path still resolves the bare head `"Future"` (type argument dropped — the #1079 mechanism) to the 4-byte default. On the base, both sides used 4 bytes: consistent, and therefore correct. The read-side fix broke the previously-consistent round-trip for 1-byte payloads.

Confirmed via `_element_mem_size` instrumentation on both commits. The `array_fold`/`array_mapi` variants fail loud at check (E130/E202) and cannot go silent.

## Fix vehicle

The write side is exactly #1079's scope (bare-head `Future` in the map-family closure-return inference); the fix ships on the open `fix/1079-map-future-elements` branch with a Bool-payload (2121-pattern) regression test pinning read/write stride agreement. Found by the adversarial delta review of the #1074 commit, which completed after #1041 merged.

### LINKED ISSUE #1082: Collection-alias Array<Future<Alias>> combinators skip payload canonicalization: silent wrong values (the #1074 residual arm)
## Summary

Array combinators over a **collection alias whose element is a Future of an alias** — `type FlagA = Bool; type Rows = Array<Future<FlagA>>;` through `array_concat` / `array_reverse` (and the slice/filter siblings) — return **silently wrong values** on current `main`: concat of `[F, T]` with itself reads back `2122` (expected `2121`); reverse reads back `2222`. The Int-payload variant regressed from a loud trap to a **silent 0**. Check and verify green, exit 0.

Partially a regression from the #1041 merge (the concat shape was correct on the pre-merge base; reverse was a loud skip; the Int shape was a loud trap — all three are now silent-wrong or wrong). Released v0.1.4 is NOT affected.

## Repro (verified as a base-vs-post differential)

`cc_rows_bool_concat.vera` and siblings (adversarial-review scratchpad, `advF-1074/`): build `[F, T]` as `Array<Future<Bool>>`, rebind through `@Rows`, then `array_concat(@Rows.0, @Rows.0)` — base 2121, current main 2122.

## Root cause

`_infer_concat_elem_type`'s **collection-alias arm** (vera/wasm/calls.py:990 region, the #1064 arm) canonicalizes the collection name, extracts the element spelling — and when that element is a **compound** (`Future<FlagA>`), returns it **verbatim**, skipping the payload canonicalization the sibling bare-element branch applies. `_strip_future` then hands the bare alias `FlagA` to the size dict, which defaults to 4 bytes for a 1-byte payload. Traced in-process: `_infer_concat_elem_type(@Rows)` → `'Future<FlagA>'` → `_strip_future` → `'FlagA'` → `_element_mem_size` = 4 (want 1).

This is the one decider path the #1074 fix missed — its CHANGELOG bullet's claim that the four deciders route through the payload canonicalizer "uniformly" is wrong for this arm; the fix commit corrects that wording. The direct `Array<Future<FlagA>>` first arm and the bare hidden-alias variant are correctly fixed; only the collection-alias-of-Future-element path is broken.

Note for the fix: the `release/v0.1.5` branch carries the #1061-merged restructuring of this function (single-exit canonicalization wrapper over a raw inference) — whether the wrapper's exit canonicalization already covers this path on that branch must be tested first; on `main` as merged it is broken as described.

## Fix vehicle

Folds into the open `fix/1079-map-future-elements` branch (same Future-payload element-sizing family). Found by the adversarial delta review of the #1074 commit, which completed after #1041 merged.

aallan:
Three bugs in the Future-element sizing family, all check-green:

- **#1079** — `array_map`/`array_mapi`/`array_fold` over `Array<Future<T>>` mis-sized the output element (or fold accumulator): the closure-return inference returned the bare head `Future` with the type argument dropped, so i64/pair payloads got the 4-byte default — a WASM-validation trap behind an exit-0 compile.
- **#1081** — the read/write stride desync that regression: `array_map` over `Array<Future<Bool>>` (direct spelling, identity pure closure) silently misread every element on main after the #1041 read-side fix — the write side still used the bare-head default. Base-correct 2121-pattern values are pinned.
- **#1082** — the collection-alias arm (`type Rows = Array<Future<FlagA>>`) returned a compound element spelling verbatim, skipping payload canonicalization — silent wrong values through concat/reverse. The CHANGELOG's "uniformly" claim from #1074 is corrected accordingly.

The fix renders full compound spellings through the #1074-extended payload canonicalizer at the closure-return/accumulator inference sites, aligning write strides with the fixed read side.

Closes #1079
Closes #1081
Closes #1082


aallan:
## Adversarial review — verdict: claims hold, subsumption sound, no blocking findings

Independent adversarial review (differential: fix content `443d5534` vs base `5bce347a`; the subsequent rebase to `e958dcba` changed only documentation merges — the code delta is identical).

**The #1082 subsumption — the PR's riskiest claim — is confirmed sound:** `git grep` shows exactly one caller of `_infer_concat_elem_type_raw` (the canonicalizing wrapper itself), every recursive combinator arm (concat/append/slice/filter/reverse/sort_by/flatten) routes through the public wrapper's single-exit canonicalization, and `git diff 5bce347a..443d5534 -- vera/wasm/calls.py` is empty. All six #1082 regression tests pass on the BASE (as subsumption implies — the #1067 wrapper already fixed the behavior there), and neutering the exit canonicalization flips exactly the four claimed tests RED, leaving the coincidence-green slice shape and the bare-element control green.

**Differential battery (18 probes):** every base-wrong case is fixed — traps (`Future<Int/Nat/Float64/String>` map/mapi/fold, aliased element/payload/collection spellings, chained `concat(map(...))`, the fold block-closure SlotRef-init fallback) now compute correct values, and the #1081 silent-wrong `Future<Bool>` identity map returns the base-correct 2121 pattern (also verified over 60 elements under `VERA_EAGER_GC=1` — shadow stack balanced). Five controls (plain `Int`/`String` maps, `Option<Int>` ADT, `Array<Int>` pair elements, the already-correct `Future<Bool>` fold) are byte-identical to base.

**Newly-reached compound spellings** were attacked for bare-head-consumer breakage and refuted: the `b_is_adt`/`u_is_adt` classifiers stay synchronized with the closure epilogue's `ret_is_pointer` (both sides classify the same spelling the same way, so shadow-stack pushes and pops remain balanced), and for non-Future compound heads the head-only classification is identical to base.

**Test genuineness confirmed on base:** the #1079 class fails via trap, #1081 via `assert 2222 == 2121`, controls pass. Docs verified accurate (three CHANGELOG bullets with correct failure-mode splits and the #1082 bullet crediting the #1067 wrapper; TESTING/ROADMAP counts match the collection). Gates on the fix: `test_codegen_arrays.py` 221 passed, the broader 16-file map/fold/mapi surface 1,013 passed, mypy clean.

**Non-blocking observation (pre-existing, identical on base):** for 1-byte `Future<Bool>`/`Future<Byte>` map outputs, both array_map and the closure epilogue classify the payload as a heap pointer and push/pop a transient shadow-stack root holding the scalar — spurious but balanced and lockstep on both sides, no correctness or GC impact. Not a defect of this PR.

Review disposition: no findings to resolve. With CI green (26/26 at `e958dcba`) and the CodeRabbit ledger clean, this PR is converged.

PR #1101, opened 2026-07-17
Fix skipped-callee codegen diagnostics

### LINKED ISSUE #1100: Caller of an E602-skipped function emits a raw `unknown func` WAT error instead of a clean diagnostic
When codegen skips a function with `[E602]` (the function is dropped from the emitted module), any **caller** of that function still emits `call $f` / `return_call $f` to the now-absent symbol, so the whole module fails to assemble with a raw wasmtime error — `WAT compilation failed: unknown func: failed to find name $f` — instead of a clean compiler diagnostic naming the real cause (the skipped callee).

## Repro

Check-clean and verify-clean; fails at `run`/`compile`:

```vera
private fn tally(-> @Int) requires(true) ensures(true) effects(pure) {
  let @Map<String, Array<Int>> = map_insert(map_new(), "a", [1, 2]);
  map_size(@Map<String, Array<Int>>.0)
}

public fn main(-> @Int) requires(true) ensures(true) effects(pure) {
  tally()
}
```

- `vera check` / `vera verify` → exit 0.
- `vera run` / `vera compile` → `WAT compilation failed: unknown func: failed to find name $tally`.

`tally` is E602-skipped because a `Map` with an `Array`-typed value is an unsupported host-import shape (a loud, sound skip — not a silent miscompile). But the E602 skip drops only the callee; `main`'s `return_call $tally` is still emitted, so the module cannot assemble.

## Impact

Loud, not silent — the program never runs with a wrong answer, so this is a diagnostics-quality gap, not a soundness hole. But the surfaced error is a raw WAT/wasmtime message pointing at generated output, not a source-located compiler diagnostic, so the author cannot tell why the program failed or which construct is unsupported.

## Path forward

When a function is E602-skipped, either propagate the skip to its (direct and transitive) callers with a clear diagnostic, or have call-site emission detect that the resolved callee was dropped and raise a source-located `[E602]`-family diagnostic naming the skipped callee. Any `[E602]`-skipped callee reproduces it, not just the `Map`/`Set` shape above.

Found in the #1098 adversarial review (the v0.1.5 #1097 fix), where the negative test sidestepped it by putting the skipped construct directly in `main`.

pollychen-lab:
## Summary

- track functions that codegen registers but later skips before emitting WAT
- turn later calls to those skipped callees into a source-located E602 diagnostic
- update the CLI regression so skipped-callee callers no longer surface raw `unknown func` WAT errors

Fixes #1100.

## Validation

- `uv run pytest tests/test_cli.py -q`
- `uv run pytest tests/test_codegen_calls.py tests/test_codegen_infrastructure.py tests/test_codegen_monomorphize.py -q`
- `uv run pytest tests/test_codegen_effects.py tests/test_codegen_host_effects.py -q`
- `uv run ruff check vera/codegen/core.py vera/codegen/functions.py vera/wasm/calls.py vera/wasm/context.py tests/test_cli.py`
- `uv run mypy vera/codegen/core.py vera/codegen/functions.py vera/wasm/calls.py vera/wasm/context.py`
- `git diff --check`
- direct #1100 repro now exits with source-level E602 diagnostics and no `unknown func`
- `rg -n "Karim13014|claude-code@anthropic.com" .git .`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler diagnostics when a function is skipped during code generation.
  * Errors now identify the original call site with a clear E602 message.
  * Prevented confusing backend “unknown func” failures in affected compilation scenarios.
  * Standardised E602 reporting across text and JSON error output.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -739,10 +739,11 @@ class TestCmdCompile:
         err = capsys.readouterr().err
         assert len(err) > 0
 
-    # A CodegenSkip drops a function with an explanatory E602 warning; if that
-    # function has a CALLER, the caller's `call $f` then dangles and WAT
-    # assembly fails with an opaque `unknown func`.  `h` skips (`hash(@Decimal)`
-    # is unsupported in codegen), `main` calls it.  The program is check-green,
+    # A CodegenSkip drops a function with an explanatory E602 warning. If that
+    # function has a CALLER, the caller must now fail at the call site with a
+    # second E602 instead of emitting a dangling `call $f` and letting WAT
+    # assembly fail with an opaque `unknown func`. `h` skips (`hash(@Decimal)`
+    # is unsupported in codegen), `main` calls it. The program is check-green,
     # so this exercises the codegen-error branch that type/syntax errors return
     # before reaching.
     _SKIP_WITH_CALLER = (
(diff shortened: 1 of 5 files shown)


aallan:
Thanks for the PR. I'm going to decline this one, though, because the fix depends on definition order and leaves the reported bug reproducible.

`_skipped_fns` is populated as functions are lowered in source order, then checked at each call site. When a caller is defined *before* the skipped callee — a forward reference — the set is still empty at the point the call is lowered, so `return_call $f` is emitted to an absent symbol and you get the same `unknown func: failed to find name $f` error. That's the exact shape in #1100's own repro (`main()` sits above `tally()`), so the bug still reproduces with this applied. A robust fix computes the skip set across the whole module in one pass *before* lowering any calls, so it's independent of definition order. (CI would also stop this on the missing `CHANGELOG.md` entry — `scripts/check_changelog_updated.py` gates changes under `vera/`.)

#1100 is already tracked as a bug and is on our near-term list, so we'll land the two-pass version there. Thanks for taking a look.

PR #1104, opened 2026-07-17
Bundle VS Code extension for Marketplace upload

aallan:
## Summary

- bundle `extension.js` and `vscode-languageclient` into a minified CommonJS Node bundle with esbuild
- keep `vscode` external and point the extension manifest at `dist/extension.js`
- preserve `vscode-languageclient`'s executable `terminateProcess.sh` beside the bundle so stop/restart fallback remains functional
- build automatically before `vsce package`, `vsce publish`, and `vsce ls`
- exclude raw `node_modules`, source/build files, lockfiles, and local metadata from the VSIX
- retain version `0.2.0` because the rejected Marketplace upload was never published

## Why

The Marketplace rejected the first VSIX with "Your extension has suspicious content." That archive shipped the complete production `node_modules` tree: 325 files and 484,902 bytes in total.

Microsoft's current [bundling guidance](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#publishing) recommends bundling extension dependencies and excluding `node_modules`. Microsoft staff also identified raw `node_modules` as the likely cause of the [same Marketplace rejection](https://learn.microsoft.com/en-au/answers/questions/5891990/publisher-extension-uploading-issue-in-vs-marketpl).

The replacement VSIX contains 11 files, is 99.8 KB, and has no `node_modules` directory.

## Validation

- `npm ci` (0 vulnerabilities)
- `npm run check`
- `npm run build`
- `npm run package:list`
- `npm run package`
- inspected the VSIX archive: 11 files, no `node_modules`, bundled entry point present, shutdown helper mode preserved as `0755`
- installed `veralang.vera-language@0.2.0` from the generated VSIX
- fresh isolated VS Code profile activated the packaged extension on a `.vera` file
- packaged LSP published diagnostics including `E206` and `E183`
- invoked `Vera: Restart Language Server` through the VS Code command API; the packaged extension stopped and restarted the server successfully
- executed the packaged shutdown helper against a disposable process
- applicable pre-commit hooks
- `python scripts/check_changelog_updated.py`
- `git diff --check`

### DIFF
--- a/editors/vscode/.vscodeignore
+++ b/editors/vscode/.vscodeignore
@@ -1,7 +1,11 @@
 .git/**
 .gitignore
+.DS_Store
 .vscode/**
 *.vsix
+esbuild.js
+extension.js
+node_modules/**
 package-lock.json
 *.md
 !README.md
--- a/editors/vscode/CHANGELOG.md
+++ b/editors/vscode/CHANGELOG.md
@@ -5,6 +5,7 @@
 Language server integration.
 
 - First VS Code Marketplace release as `veralang.vera-language`
+- Bundles the extension runtime instead of shipping raw `node_modules`
 - The extension now starts Vera's language server (`vera lsp`) for
   `.vera` files: proof-aware diagnostics with verification-tier hints,
   expression-type hover, De Bruijn slot go-to-definition, and
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -94,21 +94,21 @@ code --install-extension vera-language-0.2.0.vsix
 
 ```bash
 git clone https://github.com/aallan/vera.git
-cd vera/editors/vscode && npm install && cd -
+cd vera/editors/vscode && npm install && npm run build && cd -
 ln -s "$(pwd)/vera/editors/vscode" ~/.vscode/extensions/vera-language
 ```
 
 **Existing clone** (run from the repo root):
 
 ```bash
-(cd editors/vscode && npm install)
+(cd editors/vscode && npm install && npm run build)
 ln -s "$(pwd)/editors/vscode" ~/.vscode/extensions/vera-language
 ```
 
 Then reload VS Code. Any `.vera` file will be recognised automatically.
-The `npm install` fetches the LSP client library
-(`vscode-languageclient`); skipping it is fine — you just get syntax
-highlighting without the language server.
+The build bundles the LSP client library into `dist/extension.js`.
+Skipping it leaves the declarative syntax highlighting available, but
+the language server integration cannot start.
 
 ## Scope reference
 
--- a/editors/vscode/esbuild.js
+++ b/editors/vscode/esbuild.js
@@ -1,0 +1,39 @@
+"use strict";
+
+const fs = require("node:fs");
+const path = require("node:path");
+const esbuild = require("esbuild");
+
+const extensionRoot = __dirname;
+const distDir = path.join(extensionRoot, "dist");
+const terminateProcess = require.resolve(
+    "vscode-languageclient/lib/node/terminateProcess.sh",
+);
+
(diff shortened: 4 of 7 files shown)


aallan:
## Review — sound fix, ready to merge and re-attempt upload (no blockers)

Reviewed the full diff plus the packaged VSIX contents from the green `Package VSIX` run. The bundling approach is correct and well-executed; the findings below are advisory.

### The fix targets a plausible cause and follows Microsoft's own guidance
The rejected VSIX shipped the full 325-file production `node_modules` tree. Bundling with esbuild and excluding `node_modules` is exactly [Microsoft's recommended path](https://code.visualstudio.com/api/working-with-extensions/bundling-extension), and raw `node_modules` is a documented trigger for this rejection. Good hypothesis, well-supported.

### Verified
- **esbuild config is correct** — `external: ["vscode"]` (host-provided, must not be bundled), `platform: node`, `format: cjs`, `target: node16`, `minify`, `legalComments: "eof"` (preserves dependency license text).
- **VSIX contents (from `vsce ls --tree`)** — 11 files, 99.92 KB, containing `dist/extension.js` (the 349 KB bundle) and `dist/terminateProcess.sh`, and **no `node_modules`**. I specifically checked that the bare `extension.js` entry in `.vscodeignore` strips only the root source file, not the `dist/` bundle — the entry point ships correctly.
- **`terminateProcess.sh` is preserved** beside the bundle at `dist/terminateProcess.sh`, so the bundled `__dirname`-relative lookup resolves. The PR's validation exercised restart + the shutdown helper from the packaged VSIX end-to-end, which is the right way to prove this.
- **`vscode:prepublish` auto-builds** before `vsce package`/`publish`, so the bundle is always fresh.
- **Supply chain** — `esbuild` pinned to exact `0.28.1`; the 27 added lockfile deps all resolve to `registry.npmjs.org` with integrity hashes. `esbuild` is a devDependency, so none of it (including the `@esbuild/*` platform binaries) ships in the extension. Trojan-source scan of the added lines is clean.
- README and CHANGELOG updated accurately (source-build now needs `npm run build`).

### The one honest caveat
"Suspicious content" is an opaque Marketplace verdict, so the only real confirmation is the re-upload passing. This is the most likely fix and best practice — but if it still rejects, the next suspects are README content / links / images, not `node_modules`. Worth going in with that expectation.

### Minor / optional (non-blocking)
- `editors/vscode/.gitignore` ignores `out/` but not `dist/`; since the build now outputs to `dist/`, add `dist/` (and the stale `out/` can go) so the built bundle can't be committed by accident.
- `vscode-languageclient` stays under `dependencies`; with bundling + `node_modules/**` excluded, moving it to `devDependencies` is the more conventional signal. Harmless as-is (vsce packaged clean) — purely cosmetic.

Neither blocks merge. Good to land and re-submit to the Marketplace.

aallan:
Follow-up on the two minor review suggestions:

- **Local `dist/` ignore:** no change. The repository-root [`.gitignore`](https://github.com/aallan/vera/blob/codex/bundle-vscode-extension/.gitignore#L10) already ignores `dist/` at every depth, and `git check-ignore` confirms it covers both generated files under `editors/vscode/dist/`. Adding the same rule to the nested file would be redundant; retaining `out/` continues to protect older/test output.
- **`vscode-languageclient` dependency classification:** no change. Keeping the source runtime import under `dependencies` is valid even though the published runtime is bundled, and it matches Microsoft's bundled [VS Code ESLint client](https://github.com/microsoft/vscode-eslint/blob/main/client/package.json), which webpack-bundles its client while retaining `vscode-languageclient` under `dependencies`. Raw `node_modules` remains excluded and the VSIX contents are clean.

Both points were verified against the current branch; neither warrants a cosmetic patch.

PR #1105, opened 2026-07-17
Fix VS Code Marketplace README links

aallan:
## Summary

- replace the VS Code extension README's repository-relative LSP documentation links with stable absolute GitHub URLs
- remove Marketplace hyperlinks that cannot resolve until the extension has completed its first publication

## Why

`vsce` rewrites repository-relative links while packaging. Because the extension lives below the repository root, the two `../../LSP_SERVER.md` links became malformed `HEAD/../../` URLs in the packaged README and returned HTTP 404. The README also linked to the extension's Marketplace page before that page existed, producing two more 404s.

This removes known-bad metadata from the VSIX before retrying the Marketplace upload. Broken ordinary links are not documented by Microsoft as a definitive cause of the opaque "suspicious content" rejection, so this is a focused corrective step rather than a claim that the scanner's root cause is proven.

## Validation

- `npm ci`
- `npm run check`
- `npm run package:list`
- `npm run package -- --out /tmp/vera-language-0.2.0-readme-fix.vsix`
- inspected `extension/readme.md` inside the resulting VSIX
- confirmed the replacement LSP documentation URL returns HTTP 200
- confirmed the VSIX remains 11 files / 99.74 KB with no `node_modules`

## Summary by CodeRabbit

- **Documentation**
  - Updated the VS Code extension README with a direct link to the Vera language server setup guide.
  - Simplified installation instructions by retaining the command-line installation option.
  - Refreshed the links section with the Vera language website and GitHub repository.
  - Retained the TextMate bundle link for continued access.

### DIFF
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -5,7 +5,7 @@ Language server integration and syntax highlighting for the [Vera programming la
 ## Features
 
 **Language server integration** — the extension starts Vera's own
-language server ([`vera lsp`](../../LSP_SERVER.md)) for `.vera` files,
+language server ([`vera lsp`](https://github.com/aallan/vera/blob/main/LSP_SERVER.md)) for `.vera` files,
 which runs the full pipeline — parse, type-check, **verify** — on every
 edit and provides:
 
@@ -20,7 +20,7 @@ edit and provides:
 
 Requires the `vera` binary with the `[lsp]` extra from PyPI (see
 [Requirements](#requirements)); without it the extension stays in
-syntax-highlighting-only mode. See [LSP_SERVER.md](../../LSP_SERVER.md)
+syntax-highlighting-only mode. See [LSP_SERVER.md](https://github.com/aallan/vera/blob/main/LSP_SERVER.md)
 for everything the server can do, including the custom methods for
 coding agents.
 
@@ -80,8 +80,6 @@ Install **Vera Language** from the Extensions view, or run:
 code --install-extension veralang.vera-language
 ```
 
-[Open Vera Language in the VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=veralang.vera-language).
-
 ### From VSIX
 
 ```bash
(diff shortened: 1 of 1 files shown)


aallan:
## Review — complete and correct, ready to merge (README-only)

Codex's diagnosis is right: the VS Code Marketplace [does not support relative links in READMEs](https://code.visualstudio.com/api/references/extension-manifest#approved-badges) and rejects them, and this README had two relative `../../LSP_SERVER.md` links. This PR is the correct, focused fix.

### Completeness audit (the thing that matters on a third attempt)
I enumerated **every** link/image target in the final README, not just the diff:

- `https://veralang.dev/` — absolute ✓
- `https://github.com/aallan/vera/blob/main/LSP_SERVER.md` (×2) — absolute ✓ (the fixed links)
- `https://github.com/aallan/vera` — absolute ✓
- `https://github.com/aallan/vera/tree/main/editors/textmate` — absolute ✓
- `#requirements` — in-page anchor ✓

**No relative links, no `http://`, no images, no raw HTML** remain. (The only `<...>` occurrences are Vera generic-type syntax like `` `@Array<String>` `` inside backtick code spans, not tags.) The two fixed targets both exist on `main` (`LSP_SERVER.md` at the repo root, `editors/textmate/`), so they resolve.

Removing the two `marketplace.visualstudio.com/items?itemName=veralang.vera-language` self-links is also correct — those pointed at the extension's own Marketplace page, which is a 404 until the extension is actually published.

### Risk
README-only; the packaged bundle, dependencies, and manifest are unchanged, so this can't regress the extension itself.

### Caveat (same as before, and worth stating on attempt three)
"Suspicious content" is an opaque verdict — the re-upload is the only real confirmation. Relative links are a *documented* Marketplace rejection cause, so this is a strong, well-founded fix. But if it **still** rejects after this, I'd stop guessing and use the "contact support" path the rejection message offers to get the specific trigger, rather than iterate blind.

Good to merge and re-submit.

Issue #1124, opened 2026-07-20
Gate the corpus on vera fmt --check


aallan:
Nothing gates `vera fmt` over the corpus, which is how #1112 and #1123 stayed invisible for so long: `examples/` and `tests/conformance/` contain **zero inline comments** between them, and no script, CI job or pre-commit hook runs `vera fmt --check`. A regression that deleted every inline comment in the language passed the entire gate.

PR #1122 added a test-level guard — a sweep asserting no corpus file loses a comment when formatted, plus a fixture carrying one in every position the emitter has — but that is not the same as gating the corpus itself on canonical form.

## Why it could not land in #1122

8 of 204 corpus files are not currently canonical:

```
examples/file_io.vera
examples/gc_pressure.vera
examples/http.vera
examples/inference.vera
examples/io_operations.vera
...
```

Verified pre-existing — `origin/main` reports the same files. The divergence is layout, not content: blank-line handling and nested-`match` collapsing (`file_io.vera` goes 21 lines to 11 with every token preserved — `write_file`, `read_file` and all three `IO.print` calls survive). So this is a formatter-vs-corpus disagreement to settle, not a comment bug.

## The work

1. Decide whether the formatter's layout or the corpus files are right — the nested-`match` collapse in particular is worth a look before blessing it.
2. Reformat the 8 files, or fix the formatter, so `vera fmt --check` is clean across all 204.
3. Add `vera fmt --check` to `scripts/`, the CI lint job, and `.pre-commit-config.yaml`.

Once that gate exists, the class of bug fixed in #1112/#1123 cannot silently return.

## Related residuals

Two annotation-label gaps, both consistent with spec 1.3 as amended (only function parameters and the return slot retain), so neither is data loss — the comments survive as ordinary comments:

- Closure (`AnonFn`) labels detach from their slots: `fn(@Int /* item */ -> @Int)` keeps `/* item */` but no longer attached to the parameter.
- `OpDecl` labels are not carried in the AST — `_attach_annotations` walks `FnDecl` only — so an `op` parameter label round-trips as a comment rather than a binding label.

Worth doing if annotation retention is ever extended to every binding site; both were deliberately out of scope for #1122.

aallan:
Measured the corpus again while reviewing [#1131](https://github.com/aallan/vera/pull/1131), and the blocker splits into two groups that want different treatment.

## Current state

| | non-canonical |
|---|---|
| `examples/` on `main` | 6 of 39 |
[...]

PR #1114, opened 2026-07-19
Fix formatter comment attachment in declarations

Sanjays2402:
## Summary

- Keep comments attached to constructors and effect/ability operations during `vera fmt` instead of hoisting them outside the declaration.
- Cover all three declaration forms with formatter regressions.

## Related Issues

Closes #1113

## Type of Change

- [ ] Specification change
- [ ] Compiler implementation
- [x] Bug fix
- [x] Tests
- [ ] Documentation

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] My changes follow the project's coding standards
- [x] I have added/updated tests as appropriate
- [x] I have updated relevant documentation
- [ ] All tests pass locally

The 132 formatter tests, Ruff, Mypy, and documentation-count gate pass. The full sandboxed suite reached 7,741 passed and 226 skipped; 13 unrelated localhost-server tests fail because the required network-denied sandbox blocks socket binding.

## Summary by CodeRabbit

* **Bug Fixes**
  * `vera fmt` now preserves comments within `data`, `effect`, and `ability` declarations.
  * Concurrent `await` expressions now handle transitively defined type aliases correctly, preventing incorrect lowering.

* **Documentation**
  * Clarified both supported language-server installation methods.
  * Updated page layout so explanatory text aligns with full-width code examples.

* **Tests**
  * Expanded formatter coverage to verify comment preservation in declaration bodies.

PR #1137, opened 2026-07-22
Fix contract comment placement

### LINKED ISSUE #1136: vera fmt relocates own-line comments out of contract blocks, detaching them from the clause they document
`vera fmt` moves a comment that occupies a line of its own inside a function's contract block down into the function body, detaching it from the clause it documents. Spec §1.8 rule 11 requires the opposite.

## Reproduction

```vera
public fn demo(@Int -> @Int)
  -- leading comment before requires
  requires(true)
  ensures(true) -- trailing comment on ensures
  -- leading comment before effects
  effects(pure)
{
  @Int.0
}
```

`vera fmt` emits:

```vera
public fn demo(@Int -> @Int)
  requires(true)
  ensures(true)  -- trailing comment on ensures
  effects(pure)
{
  -- leading comment before requires
  -- leading comment before effects
  @Int.0
}
```

Both leading comments end up in the body, above a statement they have nothing to do with. The trailing comment is handled correctly — that path was fixed by #1123.

## Scope

Narrow. Own-line comments are preserved in place everywhere else that was checked — before a `data` declaration, before a constructor inside one, before a statement in a block. Contract clauses are the only position that fails.

| Position | Behaviour |
|---|---|
| own-line before a `data` decl | preserved |
| own-line before a constructor | preserved |
| own-line before a statement | preserved |
| trailing, same line as a contract clause | preserved (#1123) |
| **own-line before a contract clause** | **relocated into the body** |

## Why the existing invariants miss it

Nothing is lost and the output is stable: comment count is identical across the transform, and a second `fmt` pass is a fixed point. So the count and idempotence invariants added in #1122 both hold. Only a *position* assertion distinguishes a comment that moved from one that did not, and the claim points added in #1122 do not cover the gap between the signature and the first contract clause.

## Cause

#1122 attaches a comment to the innermost construct whose **span contains** it. A comment sitting between the signature line and `requires(` is inside no clause's span — only the declaration's — so the declaration-level backstop claims it and emits it at the top of the body.

The general rule that is missing is *leading-comment attachment*: binding an own-line comment to the construct that **follows** it rather than the one that contains it. Trailing attachment exists; leading attachment does not. Contract clauses are simply where the absence becomes visible, because every other site happens to have containment coincide with intent.

## Spec position

§1.8 rule 11:

> A formatter MUST NOT discard a comment. One occupying a line of its own **stays on its own line, above the statement or declaration it precedes.**

The rule's fallback — "where reformatting leaves no such position, the comment moves to the end of the enclosing declaration" — does not apply here: §1.8 rule 7 keeps each contract clause on its own line, so the position the comment needs still exists after formatting.

One wording ambiguity worth resolving as part of the fix: the rule says "the statement or declaration it precedes", and a contract clause is strictly neither. Either the formatter should honour the evident intent, or the rule should name contract clauses explicitly. The intent seems clear — a comment explaining `requires` belongs beside `requires`, and moving it into the body actively misinforms a reader — but the letter should be made to match.

## Acceptance

`examples/maximum_syntax.vera` (added in #1131) is the natural regression corpus: it annotates individual contract clauses as a teaching device, which is exactly the shape that fails.

The gate should be **comment position, not file identity**. After the fix, no comment in that file may change position relative to the construct it documents. Byte-identity is the wrong bar and is not reachable — `vera fmt` also collapses the multi-line `type Percentage = { … }` refinement to one line and normalises blank lines, both of which are correct canonicalisation under §0.2.3 and should not be suppressed to satisfy a comment test.

A minimal unit test over the reproduction above belongs alongside it, asserting position rather than presence.

## Related

- #1131 — adds the example; merged with the file unformatted, deliberately, because formatting it would destroy the annotations
- #1124 — corpus `vera fmt --check` gate; this is one reason a file cannot currently be made canonical without losing comment attribution
- #1123, #1112 — the comment-system work this sits adjacent to

## Separate observation, not part of this bug

`vera fmt` also rewrites `"unicode=\u{1F600}"` to the literal `"unicode=😀"`. §1.8 says nothing about string-literal normalisation, and §1.3's design note states the escape table is "the canonical and only mechanism for embedding special characters in strings" — which reads as the escape being canonical rather than the glyph. Worth deciding separately; it damages any example whose purpose is demonstrating escape syntax.

Sanjays2402:
## Summary

Keep own-line comments attached above the `requires`, `ensures`, or `effects` clause they document instead of relocating them into the function body. Add a placement and idempotency regression.

## Related Issues

Closes #1136

## Type of Change

- [ ] Specification change
- [ ] Compiler implementation
- [x] Bug fix
- [x] Tests
- [x] Documentation

## Checklist

- [x] All tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `vera fmt` moving standalone comments into function bodies.
  * Standalone contract comments now remain directly above the corresponding `requires`, `ensures`, and `effects` clauses.
  * Formatting remains stable across repeated runs.

* **Tests**
  * Added coverage to verify correct comment placement and idempotent formatting.


aallan:
Thanks for picking this up, and thanks for #1114 — that one landed cleanly.

I'm going to close this one, for three reasons.

**1. CI is red.** The `lint` job fails on `scripts/check_doc_counts.py`:

```
ERROR: 3 stale count(s) in documentation:
  TESTING.md total tests: doc says 8281, live is 8282
  README.md project-status tests: doc says 8281, live is 8282
  ROADMAP.md: test count: doc says 8281, live is 8282
```

You updated the per-file `test_formatter.py` row in `TESTING.md` but not the three headline totals derived from it. `check_doc_counts.py` is a pre-commit hook (`.pre-commit-config.yaml:137`) and its `files:` filter matches both `TESTING.md` and `tests/test_*.py`, so this would have failed locally before the push. Setup is in CONTRIBUTING.md, "Pre-commit Hooks".

**2. The fix is incomplete.** It handles contract clauses, but the same defect survives at two other anchor positions. Running this branch against `examples/maximum_syntax.vera`:

```
-- Trailing where for mutual recursion.   ->  relocated INTO the where block, re-indented
-- `\( )` interpolates an expression...   ->  relocated OUT of the match arm to column 0
```

Both are leading-comment attachment failures — the same root cause the patch fixes for `requires`/`ensures`/`effects`. `_collect_interior_anchors` gains `FnDecl` contract and effect spans here, but still doesn't anchor `where` clauses or match arms, so a comment in either position is still claimed by the enclosing construct and re-emitted somewhere else.

**3. It doesn't meet the acceptance gate.** #1136 was filed with `examples/maximum_syntax.vera` as its gate: that file has to survive `vera fmt` byte-identical. It doesn't yet, so the issue isn't closed by this change.

In fairness to you, **the scope section in #1136 is wrong, and that's on us.** It lists `data` declarations, constructors, and statements as safe and calls the scope "narrow" — `where` clauses and match arms were never checked. I've since confirmed they fail too, and I'll correct the issue. A patch that implements exactly what the issue described was always going to come up short here.

We're taking this one ourselves rather than asking you to revise, because it's entangled with #1124 and the real fix needs the wider anchor coverage plus a position-asserting regression across the example corpus. Position is the key part: comment count and `fmt` idempotency both hold *while a comment is being moved*, which is why our existing invariants never caught any of this.

For a future PR, the local gate is the bar:

```bash
pre-commit install && pre-commit install --hook-type pre-push
pytest tests/ -q
mypy vera/
python scripts/check_conformance.py
python scripts/check_examples.py
```

That runs in a couple of minutes and would have caught reason 1 outright, and reason 2 if the change had been checked against the example corpus rather than the reproduction snippet alone. Whatever your workflow looks like, running the gate before opening is what separates a contribution from a review task — and it's the difference between this PR and #1114.

Genuinely appreciate the interest in the formatter; it's a fiddly corner of the compiler. If you want a next one, the open `bug`-labelled issues are the best place to look, and anything that ships with the gate green gets a proper review.

aallan:
One addendum, because it reinforces the same point rather than adding a new one.

CodeRabbit left an unaddressed review comment on `tests/test_formatter.py:593` at 13:51, five minutes after this opened:

> **Cover an own-line comment before `ensures`.** Line 575 only tests an inline trailer, so this regression does not prove that a comment immediately preceding the second contract remains attached to `ensures`.

It's right. The test asserts own-line placement before `requires` and before `effects`, but for `ensures` it only pins the inline trailer — so the middle clause of the three the PR description names has no placement assertion at all. The branch was never updated after that comment landed.

That is the same defect as reason 2 in my previous comment, one scale down: reason 2 is positions the *fix* doesn't anchor, this is a position the *test* doesn't assert. Both are omissions that nothing fails loudly on — the suite stays green either way, which is exactly why a formatter needs position assertions rather than count or idempotency ones.

Issue #1263, opened 2026-08-08
Inference provider registry carries model IDs that vendors no longer document, and CI cannot detect it


chethanuk:
Split out of #1261 at your suggestion.

## The problem

`_PROVIDERS` in `vera/runtime/inference.py` pins a `default_model` per provider. Two of the current entries name models their vendor no longer documents:

| Registry entry | Status |
|---|---|
| `gpt-4o-mini` (openai) | absent from [OpenAI's active model list](https://developers.openai.com/api/docs/models); the cost-tier recommendation there is now `gpt-5.6-luna` |
| `kimi-k2-0905-preview` (moonshot) | absent from [Kimi's pricing page](https://platform.kimi.ai/docs/pricing/chat), which lists Kimi K3, K2.7 Code, K2.6, and Moonshot V1 |

I have not checked `claude-haiku-4-5-20251001` or `mistral-small-latest` (the latter is a floating alias, so it may be self-maintaining).

**Delisting is not always removal from the API**, so each entry needs verifying individually rather than a batch edit. xAI, for instance, publishes a [retirement page](https://docs.x.ai/developers/migration/may-15-retirement) stating withdrawn slugs "continue to resolve" — but only for the slugs in its table. Whether an undocumented name errors, 404s, or silently aliases is vendor-specific and often undocumented.

Also stale while you are in the area: `spec/09-standard-library.md` cites `platform.moonshot.ai` for Moonshot docs, which now 301s to `platform.kimi.ai`.

## Why CI cannot catch this

Every provider test asserts the model against the registry's own literal:

```python
assert sent_body["model"] == "gpt-4o-mini"
```

That pins the value to itself. The test stays green whether or not the ID resolves at the vendor, so the suite gives no signal as these rot. This is the part I think is worth solving rather than just correcting the two names — otherwise the same drift recurs silently.

Options, roughly in increasing cost:

1. **Do nothing structural**, refresh the names, accept periodic manual audits.
2. **A scheduled advisory job** (like `limitations-sync.yml`) that checks each `default_model` against a key-free listing endpoint and files or updates a drift issue. `https://openrouter.ai/api/v1/models` covers the OpenAI-compatible providers in one unauthenticated request and carries pricing, which also makes the cheap-tier default choice evidence-based. It does not cover Anthropic's own catalogue.
3. **Prefer floating aliases** where a vendor offers them (`mistral-small-latest` already is one). Trades silent rot for silent behaviour and cost changes, so probably not right for every row.

[...]

PR #1278, opened 2026-08-11
feat(inference): add DeepSeek provider

chethanuk:
Fixes #450

## Summary
- One row in `_PROVIDERS` (`vera/runtime/inference.py`): `https://api.deepseek.com/v1/chat/completions`, `VERA_DEEPSEEK_API_KEY`, `bearer` auth, `openai` response style. No dispatch code, since `_call_inference_provider` branches only on those two fields and DeepSeek fits both existing arms.
- Appended last. Insertion order is the auto-detect precedence, so appending leaves every existing key's behaviour untouched.

## Two corrections to the issue text
Both checked against DeepSeek's own docs rather than assumed.

- `deepseek-chat` is discontinued. [DeepSeek's changelog](https://api-docs.deepseek.com/updates/) says `deepseek-chat` and `deepseek-reasoner` "will be discontinued in three months (2026-07-24)", a date now past, and [pricing](https://api-docs.deepseek.com/quick_start/pricing) documents only `deepseek-v4-flash` and `deepseek-v4-pro`. Default is `deepseek-v4-flash`, the cheap tier, matching the registry's convention. What a request carrying a discontinued name actually does is not documented, and I have no DeepSeek credential to test it, so this claims nothing about call-time behaviour.
- Thinking mode is a request field, not a model name. The issue routes R1 via `VERA_INFERENCE_MODEL=deepseek-reasoner`. On v4 it's selected by a body field, `"thinking": {"type": "enabled"|"disabled"}`, and `_call_inference_provider` sends only `model` and `messages`, so no `VERA_INFERENCE_MODEL` value can reach it. Rather than substitute a plausible-looking string, the R1 claim is removed from the spec row, the registry comment, and the docs. Sending the `thinking` field is a real feature and a separate change.

## Deviation from the issue
It asks for an env-var fetch inside `host_inference_complete`. That function already loops `_PROVIDERS.items()` and does `_env.get(cfg.env_key, "")` generically, so a per-provider fetch would be dead code. Not added. The issue also names `vera/codegen/api.py`, which moved to `vera/runtime/inference.py` in #421.

## Known gap
No test here can catch a dead model ID. Every provider test asserts `sent_body["model"] == "<the literal from the registry>"`, pinning the value to itself. Tracked in #1263.

## Test plan
- [x] Tests written before the row, RED on `ValueError: Unknown inference provider 'deepseek'. Valid values: anthropic, mistral, moonshot, openai.`
- [x] Mutation check: row deleted, the two existence tests go RED; restored, 64 pass.
- [x] `test_deepseek_provider` asserts the endpoint, `Authorization: Bearer …`, absence of `X-api-key`, the default model, absence of Anthropic's `max_tokens`, and the exact turn list `[{"role": "user", "content": "prompt"}]` rather than just that `messages` is present.
- [x] The append-last invariant is parametrized over every earlier provider, from a literal tuple. A list sliced from `_PROVIDERS` would shrink along with a relocated row and pass vacuously. These cases stay green when the row is deleted, by design: they pin position, not existence.
- [x] No API key used; `urllib.request.urlopen` is patched throughout, following `test_mistral_provider`.
- [x] `mypy vera/`, `ruff check .`, `check_doc_counts.py`, `check_site_assets.py`, `check_diagnostic_fields.py`, `check_changelog_updated.py` green. Both edited examples pass `vera fmt --check`, `check`, and `verify`.

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek as a supported inference provider.
  * Added automatic provider detection using `VERA_DEEPSEEK_API_KEY`.
  * Added explicit DeepSeek selection via `VERA_INFERENCE_PROVIDER=deepseek`.
  * Configured the default DeepSeek model and compatible authentication and response handling.

* **Documentation**
  * Updated setup guidance, provider lists, examples, roadmap, changelog and project statistics.

* **Tests**
  * Added coverage for DeepSeek requests, authentication, model selection, provider detection and precedence.


chethanuk:
@aallan Same one-row registry change as #1261, with two corrections to the issue text (the model IDs it names are discontinued, and thinking mode is a request field rather than a model name) documented in the description. All 27 applicable checks green. Ready for review and merge.

aallan:
Thanks — queued for review after the in-flight v0.1.10 release lands (`release/v0.1.10`, twelve merged PRs, final work in progress this week). Same note as the xAI PR: whether the provider registry takes a DeepSeek entry is a maintainer product call that precedes the code review, and both provider PRs will be triaged together post-release.

aallan:
# Adversarial review record — PR #1278 (four-lens fitness protocol, 2026-08-13)

Reviewed at the contributor's commit rebased onto post-#1261 `main` (the flagship registry). Verdict: **PASS with one reshape applied** — the flagship default and the test bar — landed as the maintainer-side commit `db271000`.

## Fitness verdict

| Lens | Verdict | One line |
|---|---|---|
| 0. DESIGN.md (veto) | **PASS** | No language machinery — one `_ProviderConfig` row, zero new fields, no dispatch branch: the registry's one-row promise working as designed (principle 3). Where it *did* engage DESIGN is principle 5 (contracts as source of truth): a program's `ensures` over `Inference.complete` output is written against what the default model can do, so a cheap-tier default silently weakens the premise every such contract rests on — that drove the reshape, not a reject. |
| 1. Fit | PASS | The established registry-extension pattern exactly. |
| 2. Addresses #450 | PASS | Footer and `closingIssuesReferences` agree on the single issue; the issue asks for precisely this row. |
| 3. Entirely closes #450 | PASS | Three of four specifics land; the issue's env-var-fetch suggestion was verified already-generic in the live code (dead code if added), and its file path predates the #421 module move — both documented rather than faked. |

## The reshape (`db271000`)

- **Default pinned to `deepseek-v4-pro`** — DeepSeek's flagship, verified against their live documentation, never memory: the create-chat-completion reference lists exactly two model values (`deepseek-v4-flash`, `deepseek-v4-pro`); the changelog dates V4-Pro's GA on API to 2026-08-13 against Flash's 2026-07-31 public beta; and the introduction's first sample call sends `deepseek-v4-pro` in all three languages. No chat-vs-reasoner ambiguity exists at this generation (the legacy pair was discontinued 2026-07-24; thinking mode is a request field, not a slug). The contributor's cheap-tier choice was correctly argued *against the registry convention as it stood* — that convention changed with #1261.
- **Tests raised to the #1261 bar**, each RED-proven: model and forwarded-API-key as literals (the key at its verified positional index), the Content-Type assertion the xai sibling carries, and the provider-precedence tuple widened to five — as written, a DeepSeek row relocated above `xai` could not fail it; the relocation mutation now fails exactly the xai case and no other. Six-mutant battery, all discriminating, green-mutate-restore verified.
- The false-by-append xai docstring sentence ("xAI is last in the registry") now states the invariant it actually checks.

## Rebase notes (three silent semantic conflicts caught)

Git auto-merged the registry file *wrong* — both sides appended at the same anchor and the "new providers append here" comment was duplicated; a per-file test-count bump made independently by both sides (58→64 twice) would have shipped 64 where the truth post-union is 71; and a take-one-side ROADMAP resolution would have resurrected a closed issue's row. All three resolved against the oracles, with SKILL's provider-precedence prose widened to six and two "five providers" count phrases the diff never flagged corrected.

## Left deliberately open — the endpoint question

The row uses `https://api.deepseek.com/v1/chat/completions` — the issue's own specification and the sibling rows' style — but DeepSeek's current docs give `https://api.deepseek.com` (OpenAI-format) with no `/v1` anywhere, while third-party references describe `/v1` as still accepted. Without a credential nothing here can test which is true on the wire, so the literal is pinned as-is and the question is recorded for a maintainer-side smoke test (`VERA_DEEPSEEK_API_KEY` against `examples/inference.vera`, the same live check just performed for xAI): if `/v1` fails, the fix is one string plus its pinned test.

## Gates

Full suite 10,319 passed / 140 skipped (collected 10,485 across 162 files, matching the oracle), 214 conformance, 42 examples, mypy/ruff clean, doc-counts and site-assets consistent, spec-examples green with §9.5.5 extended to the six-model flagship list. Diff vs `main` restricted to `vera/` + `tests/` is three files; the registry diff is a pure seven-line addition with all five landed flagship pins byte-intact.

*Reviewer: Opus subagent, four-lens fitness protocol, DESIGN veto first; vendor claims verified against live DeepSeek documentation with sources quoted in the commit message.*

aallan:
Closing the review record's one deliberately-open item (the endpoint question): verified working end-to-end with a live `VERA_DEEPSEEK_API_KEY` against `examples/inference.vera` — the `/v1` path is accepted on the wire despite no longer appearing in DeepSeek's docs, so the pinned `https://api.deepseek.com/v1/chat/completions` literal stands as shipped. Both providers added this cycle (xAI, DeepSeek) have now passed the same live smoke test.

Issue #1277, opened 2026-08-10
A module declaring a demand-injected prelude ADT's name subtracts it from every other namespace's member set (and the collision shape compiles to nothing)


aallan:
Found during PR #1272's convergence (CodeRabbit raised the reservation question; the verification produced this precise map). Pre-existing — behaviourally identical at `ef7c04c0` — and NOT release-gating: inert at emission today, loud where it bites.

Codegen's per-namespace ADT membership (#1253) derives global infrastructure by subtracting what the namespaces declare from the registered layouts — an order-independent rule that correctly covers the prelude ADTs injected in Pass 1.2, after Pass 0.5 computes the membership sets. The subtraction is not namespace-scoped, though: when a module declares `data Json`, `Json` enters the declared set, is removed from infrastructure, and is therefore absent from the member set of **every namespace except that module's** — including the entry program's, which never declared it and legitimately sees the prelude's. Measured: `builtin_snap_has_Json=False`, `namespace_declared=['Json']`, `members[None]` loses `Json` while `members[('jlib',)]` keeps it. The Pass-0.5 built-in snapshot unioned in as a floor does not protect these four names, because it predates their injection — the same Pass-0.5-vs-1.2 asymmetry #1253 fixed, one layer down — and no E609/E610 collision rail fires on the declaration either, those rails keying on that same snapshot.

**Inert at emission today**: `AliasEnv.data_types` changes an answer in exactly one place (`naming._resolve_named`) and only for `Decimal` and the single `REMOVED_ALIASES` entry `Float`, so no rendering and no emitted WAT moves — the whole 255-program corpus is byte-identical with and without the membership rule. A latent constraint on the next consumer of that set rather than a live defect.

**Fix direction**: scope the subtraction per namespace (a name another namespace declares should not leave THIS namespace's infrastructure), or teach Pass 0.5 which prelude ADTs the program will demand so the snapshot is complete when membership is built — **not** by reserving the names, which spec §8.4.1 forbids: the prelude's data types are ordinary public declarations a program names and shadows, and the reserved namespace is the `Vera` prefix alone (E154, #1254/#1260). Documented at `CodeGenerator._adt_members_in_scope`.

[...]

PR #1339, opened 2026-08-21
Refuse an entry-file data declaration that contends with a module's

### LINKED ISSUE #1312: Entry-file and module `data` declarations sharing a name silently drop the caller when shapes differ
When the entry file and an imported module both declare a `data` type of the same name with different shapes, the program checks green, compiles with exit 0, and silently drops the calling function from the emitted module — only warnings are reported.

## Repro

```vera
-- blib.vera
module blib;

private data Json { JBlob(Int) }

public fn probe(@Int -> @Int)
  requires(true)
  ensures(@Int.result == @Int.0)
  effects(pure)
{
  @Int.0
}
```

```vera
-- main.vera
import blib(probe);

private data Json { JMine(Int) }

public fn consume(@Json -> @Int)
  requires(true)
  ensures(true)
  effects(pure)
{
  1
}

public fn main(@Unit -> @Int)
  requires(true)
  ensures(@Int.result == 7)
  effects(pure)
{
  probe(7)
}
```

```
$ vera check   → OK (rc 0)
$ vera compile → rc 0, ok: true, warnings [E602, E620]
   exports == ['consume']   — main is silently gone
```

Identical at `release/v0.1.12` (8e08ec90) and after the #1277 fix.

## Why the #1277 rail does not catch it

The entry file's own declaration suppresses the prelude's injection, so no prelude `DataDecl` exists for the E621 rail to contend with: the contention here is **entry-declaration vs module-declaration**, a different pair from #1277's prelude-vs-module. Same disease family as #1253 (namespace membership disagreements surfacing as silent drops), same failure surface (check-green, exit-0, missing export).

## Fix direction

The same shape-based contention treatment #1277 received, applied to the entry-vs-module pair: structurally identical declarations may share a layout; differently-shaped ones get a loud codegen-phase diagnostic naming both declarations, and the compile fails instead of shipping a module without `main`. The `data_decl_shape()` predicate and the compile-stage conformance-negative machinery added by the #1277 fix are the reusable pieces.

Found by the adversarial review of the #1287/#1277 branch while verifying that E621's fix-text was truthful. The KNOWN_ISSUES row rides that PR.

AmirF194:
## Summary

Pass 0.5 (`_register_modules`) harvests an imported module's ADT layouts into the flat `_adt_layouts` map via `setdefault`. Pass 1 (`_register_all` -> `_register_data`) then registers the entry file's own `data` declarations over that map unconditionally, with no shape check. An entry-file `data` declaration sharing a name with an imported module's, but describing a different layout, silently takes the one registered slot: the module's own constructors become `unknown constructor` inside its own bodies (E602), the functions that use them are dropped behind it (E620), and the artifact compiles with exit 0, missing its entry point.

This is the sibling of #1277's prelude-versus-module contention rail (E621), which cannot see this pair: an entry-file declaration suppresses the prelude's own injection outright (spec 8.4.1), so a contention between the entry file and an imported module never reaches the point where E621 is decided.

## Fix

Reuses `data_decl_shape` (#1277) at the Pass-1 registration site: an entry declaration sharing a module's ADT name is now only allowed to register over it when the two describe the same layout. A genuine mismatch is refused with a new diagnostic, **E622**, located at the entry file's own declaration, not at the swallowed module warnings. A module restating the entry's own shape still compiles as before, since the one registered layout is correct for both.

## Related Issues

Fixes #1312

## Verification

- New test `tests/test_entry_module_adt_namespace_1312.py`: the differing-shape case fails on `main` (E602/E620 warnings only, `exports == []`) and passes on this branch (`E622`, `cmd_compile` returns 1 over a `cmd_check`-clean program). Confirmed both ways in the same container. Two control cases (restating the module's shape, and an entry that never declares the name) stay green on both.
- Full suite green: `pytest -n auto --cov=vera --cov-fail-under=80`, `ruff check .` (incl. `--select S`), `mypy vera/`.
- Repo's own lint gates: `check_changelog_updated.py`, `check_doc_counts.py`, `check_diagnostic_fields.py`, `check_e602_clean.py`, `check_conformance.py`, `check_examples.py`, `check_licenses.py`, `check_limitations_sync.py`, `uv lock --check` all clean.
- Not independently checked: the WASM-runtime behavior of a program that hits E622 at a nested/generic import depth beyond the direct case this test covers; the diagnostic path itself is exercised directly.

## Type of Change

- [ ] Specification change
- [x] Compiler implementation
- [x] Bug fix
- [x] Tests
- [x] Documentation

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] All tests pass locally

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for conflicting entry-file data declarations and imported module layouts.
  - The compiler now reports clear `E622` diagnostics and refuses invalid builds.
  - Matching declarations and cases without contention continue to compile successfully.

- **Documentation**
  - Added guidance for the new diagnostic and updated project statistics, including test totals and error-code coverage.

- **Tests**
  - Added regression coverage for conflicting, compatible and unaffected declaration scenarios.


aallan:
Thanks for taking this on. We're going to close this one rather than iterate on it here: #1312 is already queued in the v0.1.14 burndown, and we'll land it there with the conformance fixture, the spec text for the new diagnostic, and the release bookkeeping the project's conventions ask for — and the review threads on this PR went unanswered, which is where that work would have needed to happen. The issue stays open until that fix lands. Appreciate the interest in Vera.

AmirF194:
Understood, thanks for the context. Fair point on the coderabbit thread, that should have been answered here rather than left open.

Issue #1410, opened 2026-09-06
A disclosed value passed as an argument satisfies a refined parameter with no obligation anywhere


aallan:
Passing a disclosed callee's result as a **call argument** gives the receiving function a **Tier-1 `verified`** postcondition that the compiled program refutes at run time. `vera verify` exits 0.

Measured on `origin/release/v0.2.0` at `aecc1a6ad0eecfd45fbd89d0742c3a71bc24b03f`. Distinct from #1406 / #1407: this one is **spelling-independent** — the direct call, the `let`-bound call and the wrapper all escape identically, before and after a fix that makes the taint follow the value — because the missing piece is an obligation that does not exist at all, not one that the taint fails to reach.

## Repro

```vera
type PosInt = { @Int | @Int.0 > 0 };

private fn mk(@Float64 -> @Option<PosInt>)
  requires(true)
  ensures(true)
  effects(pure)
{
  Some(float_to_int(@Float64.0))
}

private fn consume(@Option<PosInt> -> @Int)
  requires(true)
  ensures(@Int.result > 0)
  effects(pure)
{
  match @Option<PosInt>.0 {
    Some(@PosInt) -> @PosInt.0,
    None -> 41
  }
}

public fn f(@Float64 -> @Int)
  requires(true)
  ensures(@Int.result > 0)
  effects(pure)
{
  consume(mk(@Float64.0))
}
```

```
$ vera verify --json arg.vera   ->  ok: true, exit 0
    mk's   refine_bind: tier3_unguarded + E506   (neither proved nor guarded)
    consume's ensures:  verified                  <- proved FROM the disclosed fact
    f's       ensures:  verified

$ vera run arg.vera --fn f -- -7.0
Error: Postcondition violation in consume(@Option<@PosInt> -> @Int)
  ensures(@Int.result > 0) failed
```

`-7` is chosen so no fallback can coincide with it: it is neither `0` nor the `None` arm's `41`.

## Root cause

The machinery is already right when the caller **constructs** the argument. Replace the call with a literal and the caller is obligated at its own construction site:

| `f`'s body | caller's obligation | `consume`'s `ensures` |
|---|---|---|
| `consume(Some(-7))` | `refine_bind` **`violated`**, E505 | `tier3`, E522 |
| `consume(mk(@Float64.0))` | **none at all** | **`verified`** |

[...]

Issue #1362, opened 2026-09-02
Handler-clause `@Nat` payload narrowing is classified tier3 ("runtime-guarded") with no guard and no E504 disclosure


aallan:
A handler-clause payload binder narrowed to `@Nat` is classified **`tier3`** — counted in `verify --json`'s `tier3_runtime`, whose meaning is "runtime-guarded" — while **no runtime guard exists at the site**, and no disclosure fires either: zero warnings, zero `tier3_unguarded`. The program then stores and returns a negative through the `@Nat` binder with no trap.

## Repro

```vera
public fn f(@Int -> @Int)
  requires(true) ensures(true) effects(pure)
{
  handle[Exn<Int>] {
    throw(@Nat) -> { nat_to_int(@Nat.0) }
  } in {
    throw(@Int.0)
  }
}
```

```
$ vera verify --json  -> ok: true, warnings: [], the clause's nat_bind status "tier3"
$ vera run --fn f -- -7  -> -7   (no trap)
```

## Why this is its own defect and not #754's

The missing **guard** is squarely the #754/#757/#765 residual family (a `@Nat` narrowing site codegen does not yet guard). But that family's documented contract (the KNOWN_ISSUES four-site row) is that such residuals are **disclosed statically as an E504 warning** / `tier3_unguarded` — the honest "nothing guards this" bucket. This site takes the other path: the obligation is classified plain `tier3`, i.e. counted among the runtime-guarded, so both halves of the accounting are wrong at once — the summary overcounts guarded obligations, and the disclosure machinery that exists precisely for unguarded residuals never engages. Corpus-wide there are currently **0 E504 warnings and 0 `tier3_unguarded` obligations**, so this disclosure path is entirely unexercised by the suite.

The classification defect is fixable independently of the guard: the clause-binder site should either carry a real guard (#754's work) or be classified `tier3_unguarded` + E504 like its documented siblings — never plain `tier3`.

Measured identical before and after PR #1361's fixes (`b4256509` and its branch), which is where the probe that found it ran.

## Fix direction

Locate where the handler-clause binder's `nat_bind` obligation is bucketed after an unknown/undecided solver result, and route it through the same guarded-versus-unguarded classification the other narrowing sites use (the E504/`tier3_unguarded` machinery), keyed on whether codegen actually plants a guard at that site — the obligation↔guard parity discipline of #1205.

aallan:
[...]

PR #1435, opened 2026-09-07
Recover a generic constructor field's instantiation, and stop a let claiming a guard

aallan:
CodeRabbit's pass on #1420's rebased head landed four findings after that PR had already merged. All four were valid; two are code defects now sitting on `release/v0.2.0`.

## The two code defects

**A generic constructor field raised no obligation at all.** The registry's declared field type for `data Box<T> { MkBox(T) }` is a `TypeVar`, and the nested-refinement gate finds no refinement in one — so the site was skipped entirely:

```vera
private data Box<T> { MkBox(T) }

public fn f(@Float64 -> @Int)
  requires(true) ensures(@Int.result > 0) effects(pure)
{ consume(MkBox(mkint(@Float64.0))) }      -- consume takes @Box<Option<PosInt>>
```

```
verify --json  ->  ok: true, zero refine_bind obligations
run --fn f -- -7.0  ->  Postcondition violation in consume
```

The CONCRETE spelling — `data Box { MkBox(Option<PosInt>) }` — is refuted, so the two differed only in whether the field was generic. It now goes through `_nested_refinement_formal`, the same instantiated-target recovery (#747) every other binding-obligation target makes, and the shape is `violated`/E505.

**A `let` claimed a boundary guard it does not have.** Codegen's component decomposition runs at parameter entry and return exit; a `let` is neither. Deriving the flag from the target type made `let @Tuple<PosInt, Int>` record the guarded `tier3` for a runtime check nothing at that site emits — the #1362 shape — and disagreed with the refined arm three lines above it, which has always passed `guarded=False`. Now `guarded=False`.

Observing this needed a **disclosed** producer: a refutable value lands on `violated`, where the flag is never read, so a cell built on one proves nothing about it.

## The two test/doc findings

- `test_a_return_that_relabels_an_unrefined_payload_is_refuted` and `test_a_closure_argument_is_obligated` asserted only `returncode != 0`, which a compile refusal also satisfies — neither could tell a refuted contract from a program that never ran. Both now assert the postcondition violation by name.
- `TESTING.md`'s "Contract verification" row still read **411 of 533** while #1420 moved the examples to **413 of 535** — the number that PR's own corpus pin asserts. Updated (77.2%).

## Evidence

Two new cells, both with the run differential, and both mutation-validated: reverting the field to the unsubstituted registry type turns the generic cell red, and restoring the derived flag at the `let` turns the guardedness cell red. The full battery is 21 mutations across 22 naming cells, all RED.

Corpus differential against `f23db8f7` is unchanged from #1420's: the same two movers (`examples/ephemeris.vera` +2 `verified`, one state-handler probe +1 `tier3_unguarded`), and the accounting identity holds on all 370 verifiable programs.

Follow-up to #1420 (merged at `ab758f69`); relates to #1410.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generic constructor fields now correctly recover their concrete refinements during verification.
  - `let`-bound nested refinements are correctly reported as unguarded when no runtime boundary guard applies.
  - Improved handling of refinements in call-argument payloads, closures, and tuple bindings.

- **Tests**
  - Added regression coverage for postcondition failures, closure arguments, generic fields, and `let`-bound refinements.
  - The test suite now includes two additional tests.

- **Documentation**
  - Updated project, roadmap, FAQ, testing, and verifier documentation to reflect revised test totals and coverage.


aallan (commented):
Reviewed `1aafdb60` in a fresh worktree, red-first against the merged tip `35ff4def`. Both defects reproduce there and both are closed here; one low note and one nit, no soundness findings.

## Defect 1 — the generic constructor field

| Shape | `35ff4def` | `1aafdb60` |
|---|---|---|
| `data Box<T> { MkBox(T) }` built at `Box<Option<PosInt>>` | `ok: true`, exit 0, **zero** `refine_bind`, run refutes `consume_b`'s postcondition | `violated`/E505 at `constructor field`, exit 1 |
| concrete control `data BoxC { MkBoxC(Option<PosInt>) }` | `violated`/E505 | `violated`/E505 |
| two type parameters, `Pair<Option<PosInt>, Tuple<PosInt, Int>>` | silent | **two** E505s, one per field, each naming its own type |
| nested generic, `Box<Box<Option<PosInt>>>` | silent | one E505 at the innermost construction |

The asymmetry the PR describes is exactly what I measured: the concrete spelling was refuted on the merged tip while the generic spelling of the same program raised nothing at all. On the nested case only one obligation appears, which is right — the outer field's argument is itself a `MkBox(...)`, so `_all_leaves_construct` excludes it and its own site carries the claim.

Reverting the `_nested_refinement_formal` recovery turns four cells red, including the new `test_a_generic_constructor_field_recovers_its_instantiation`.

## Defect 2 — the `let`'s guard claim

Status moves exactly where it should and nowhere else. With a disclosed producer (the only way to reach the branch where the flag is read), `let @Tuple<PosInt, Int>` records a guarded `tier3` on `35ff4def` and `tier3_unguarded` here. Every other refined `let` shape — `Option<PosInt>`, `Array<PosInt>`, `Result<PosInt, String>`, a two-field user ADT, `Box<Option<PosInt>>`, `Map<String, PosInt>` — is `tier3_unguarded` on both sides, so the change is as narrow as claimed.

Checked against the artifact rather than against itself: a program whose only refined position is `let @Tuple<PosInt, Int> = prod(x)`, with `prod` returning an unrefined `Tuple<Int, Int>` so no boundary guard can mask the site, compiles to a module containing **zero** refinement guards — `grep -c 'tuple component'` and `grep -c 'Refinement violation'` over `compile --wat` are both 0 — and `vera run --fn f -- -7.0` does not trap at the `let`, running on to fail the postcondition instead. So the old `guarded=True` was a runtime-check promise nothing keeps, and `False` agrees with what codegen emits.

Letting the `let` derive its flag again turns `test_a_let_binding_never_claims_a_boundary_guard` red.

## The TESTING.md correction is a real one

The contract-verification row moves `411 of 533 (77.1%)` to `413 of 535 (77.2%)`. Independently derived across the 43 examples, the true figures are 413 Tier 1, 122 Tier 3, 535 total — and they are the same on `35ff4def` as here, so #1420 shipped a stale row and this is the correction, not a consequence of these two fixes. Test counts (13,307 across 199 files) match the collected count and every other doc that carries them.

## Corpus and gates

`35ff4def` vs `1aafdb60` over 474 programs: **zero** movers, no exit-status change, and the accounting identity holds on all 370 that emit a verification envelope. That is the expected result — neither defect's shape occurs in the corpus, which is worth saying out loud, because it means the corpus is a weak instrument for this change and the probe matrix is carrying it.

pytest 13,095 passed / 186 skipped / 26 deselected, matching the updated TESTING.md row exactly; mypy clean; ruff clean including `--select S`; conformance 250/250; 43/43 examples; `check_diagnostic_fields`, `check_site_assets`, `check_limitations_sync` all pass; `tests/test_verifier_call_arg_payload_1410.py` 51/51. The two cells that previously asserted only `returncode != 0` now assert the diagnostic text as well.

## Two notes

**N1 (low) — comment rot in the edited block.** The inserted clause severed a sentence. `vera/verifier.py` around line 5605 now reads:

> The same recovery every other binding-obligation target makes (#747) — `data Box { MkBox(Option<PosInt>) }` — is claimed by none of the arms above, which read the field's HEAD.

The subject of "is claimed by none of the arms above" used to be the FIELD; it is now "The same recovery", which is not a thing the arms above claim. A sentence boundary after `(#747)` and a fresh subject for the second half restores it.

**N2 (nit) — line length.** `vera/verifier.py:5606` is 90 characters where every other line in that comment block is 81 or shorter. Ruff has no line-length rule configured, so nothing catches it; it is only inconsistent with its neighbours.

VERDICT: CLOSED

Issue #1442, opened 2026-09-07
validate_handler accepts a Bool -> Bool handle when both prelude layouts are present, and executing it crashes the host


aallan:
`vera.runtime.server.validate_handler` accepts a `handle` whose parameter and result are not `Request -> Response`, provided the `Request` and `Response` layouts exist *somewhere* in the program and `handle` takes one WASM `i32`. `Bool` and every pointer-backed type also lower to `i32`, and the return type is not checked at all. Executing the accepted handler reinterprets its result as a `Response` pointer and crashes the host process.

**Reproduce** (check-clean; compiles)

```vera
public fn keep(@Request -> @Response)
  requires(true) ensures(true) effects(pure)
{ Response(200, map_new(), "ok") }
public fn handle(@Bool -> @Bool)
  requires(true) ensures(true) effects(pure)
{ @Bool.0 }
```

`keep` makes both layouts available; `handle`'s recorded ABI is `['i32']`; `validate_handler` returns without a diagnostic. Executing `handle` with an `HttpRequestData` in a disposable process ends with `Fatal Python error: Bus error` (SIGBUS; SIGSEGV on other runs):

```text
heap.py:155  _read_bytes_at
heap.py:575  _decode_map
heap.py:844  _decode_attrs
heap.py:1371 decode_response_adt
codegen/api.py:1448 execute
```

**Expected.** `vera serve` and the WASI server-world validator refuse a handler whose Vera parameter and return types are not the prelude `Request` and `Response` (ownership-aware identity, since #1409/#1419 key prelude types per owner), with a diagnostic rather than a crash. Independently, `_read_i32_at` and `_read_bytes_at` bounds-check the raw pointer read the way the string reader already does, so a malformed pointer or length is a structured runtime error inside the existing boundary, never a native signal.

**Test shape.** Wrong `i32` parameter with both layouts present; correct `Request` parameter with a wrong return; an unrelated ADT sharing the ABI; a valid handler control; native serving and the shared WASI validator; malformed-pointer and malformed-length decodes run in subprocesses so a native crash is a test failure, not a lost worker. The existing wrong-signature tests use `Int -> Int`, which lowers to `i64` and does not exercise this predicate.

Where: `vera/runtime/server.py` (`validate_handler`), `vera/runtime/heap.py` (raw reads), `vera/codegen/api.py` (`CompileResult.fn_param_types` holds lowered types only).

[...]

Issue #1145, opened 2026-07-23
SIGBUS: _read_wasm_string reads WASM memory out of bounds via unchecked ctypes pointer slice


aallan:
## Summary

`_read_wasm_string` in `vera/runtime/heap.py` slices WASM linear memory through a raw `ctypes` pointer **without bounds-checking `(ptr, length)`**. Because `ctypes` pointer slicing performs no bounds checking of its own, an out-of-range pair reads past the end of the WASM memory region and into wasmtime's guard page, killing the process with `SIGBUS` / `KERN_PROTECTION_FAILURE`.

Its sibling `_read_string_export`, twenty lines below in the same file, *does* perform exactly this check.

Observed 14 process deaths in a ~2-minute window while running the vera-bench test suite against `f6f586b0` (v0.1.6).

## Crash signature

Identical across all inspected reports (`~/Library/Logs/DiagnosticReports/Python-*.ips`):

```
exception : {'type': 'EXC_BAD_ACCESS', 'signal': 'SIGBUS',
             'subtype': 'KERN_PROTECTION_FAILURE at 0x0000000d9a020000'}

faulting thread:
  _ctypes.cpython-314-darwin.so   u8_get
  _ctypes.cpython-314-darwin.so   Pointer_subscript
  Python                          _PyEval_EvalFrameDefault
  ...
usedImages: _libwasmtime.dylib  (loaded)
```

`Pointer_subscript` → `u8_get` is a `ctypes` pointer **slice read**, which is precisely `buf[ptr:ptr + length]` below.

## Root cause

`vera/runtime/heap.py`:

```python
def _slice_and_decode(memory, context, ptr: int, length: int) -> str:
    buf = memory.data_ptr(context)              # raw ctypes POINTER(c_uint8)
    return safe_utf8_decode(bytes(buf[ptr:ptr + length]))   # <-- unbounded read


def _read_wasm_string(caller: wasmtime.Caller, ptr: int, length: int) -> str:
    memory = caller["memory"]
    assert isinstance(memory, wasmtime.Memory)
    return _slice_and_decode(memory, caller, ptr, length)   # no validation


def _read_string_export(memory, store, ptr: int, length: int) -> str | None:
    if length < 0:
        return None
    mem_size = memory.data_len(store)
    if not (0 <= ptr and ptr + length <= mem_size):          # <-- the check
        return None
    return _slice_and_decode(memory, store, ptr, length)
```

[...]

PR #1456, opened 2026-09-07
Check the handler by its Vera types, not by its ABI

aallan:
`validate_handler` is the one guard in front of both serving surfaces — `vera serve` and `--target wasi-p2 --world server` — and it established neither half of what the host goes on to marshal.

## What it actually asked

Two proxy questions: are `Request` and `Response` in `adt_layouts`, and does `handle` lower to `["i32"]`. Its docstring asserted these were sufficient — *"The layouts check doubles as the type check"* — and both are satisfiable by a program whose `handle` has neither type:

- the prelude injects those ADTs when the **program** mentions them, so any *other* function can supply the layouts on `handle`'s behalf; and
- `i32` is equally the lowered shape of `Bool`, of every heap ADT, and of every boxed value, so the ABI cannot distinguish a `Request` pointer from anything else that is pointer-shaped.

So this passed:

```vera
public fn keep(@Request -> @Response)
{ Response(200, map_new(), "ok") }

public fn handle(@Bool -> @Bool)
{ @Bool.0 }
```

The host then built a `Request` in guest memory, handed the pointer to a `Bool` parameter, and decoded whatever came back as a `Response` — reading a garbage headers pointer past the end of linear memory into wasmtime's guard page. That is a **SIGBUS in the host process**: not a trap, not an exception, a dead interpreter.

The existing wrong-signature test uses `Int -> Int`, which lowers to `i64` and so fails the old check for a different reason. Nothing exercised this predicate.

## What it asks now

The **declared Vera types**, from a new `CompileResult.fn_adt_signatures`, populated by `_declared_adt_name` through the same `vera.naming.classify_named` spine `_return_type_is_string` takes — so aliases resolve and refinements are transparent. It returns `None` rather than a spelling for a non-ADT, which is what makes the consumer fail closed.

It is also **ownership-aware**, and has to be. An entry file may declare its own `data Request`, which shadows the prelude's (§8.4.1) and takes the single `adt_layouts` slot for that name. Measured, that program's Vera signature is *byte-identical* to a valid handler's:

| program | `fn_adt_signatures["handle"]` | `prelude_injected_adts` |
|---|---|---|
| valid handler | `(('Request',), 'Response')` | `Request`, `Response`, … |
| entry-declared `data Request` | `(('Request',), 'Response')` | `Response`, … (**no** `Request`) |

Names alone genuinely cannot separate them; only *who declared the type* can, which is what per-owner identity says a type **is**. `CompileResult.prelude_injected_adts` records which ADTs the injection actually supplied — observed at the Pass-1.2 walk that is already filtered to injected declarations, rather than by re-deriving `inject_prelude`'s demand predicates, so prelude provenance has one source of truth and not two that drift.

The refusal names *which* of the three things is wrong, because the remedies differ — a non-data type and a differently-named type are signatures to change, while the right name declared by this program is a shadowing to delete. A message mentioning only `@Request` could not convey the last one, since the program plainly does say `@Request`.

## Containment beneath it

`_read_i32_at` and `_read_bytes_at` were the last two raw readers with no bounds check (`_read_wasm_string` grew one under #1145), and both are what `decode_response_adt` uses for the status, the headers pointer, and the body `(ptr, len)`. A shared `_require_readable` now raises the `out of bounds memory access` reason `execute()` already classifies as a clean trap — so the next gap in this family is a diagnostic rather than a dead process.

## Evidence

- **Red-first.** Every identity cell failed as `DID NOT RAISE` and every guard cell as a fault (7 × SIGBUS `-10`, 1 silently returning wrong bytes) before the change.
- **Mutation-validated.** Twelve mutations across the new guards — both reader call sites, all three bounds conjuncts, both halves of the identity test, each of the three `validate_handler` checks, `_declared_adt_name` failing open, and `prelude_injected_adts` recording unconditionally. **All twelve red, no survivors.** The shape is informative: dropping the *name* half reds 10 cells, dropping the *ownership* half reds exactly 2 — the two conjuncts guard genuinely different programs.
- One case the rewrite exposed and handles explicitly: `handle` in `exports` with no entry in `fn_adt_signatures`. That cannot happen for a program codegen accepted — the map is built from the declarations codegen compiles — but routing it through the arity branch would have told the author their handler "takes 0 parameter(s)", which is false and unactionable. It now refuses as an internal error naming what it observed. Mutation-validated: reverting to the `.get()` default reds that cell.
- **Corpus differential: 0 movers** (253 identical, 47 vacuous). The change is codegen-neutral.
- Full suite, mypy, ruff (+`--select S`), conformance, examples, corpus canonical, doc counts, site assets, limitations sync, encoding, diagnostic fields — all green.

### A note on how the tests observe this

**Nothing in the suite reproduces the crash.** The cells assert the *guard* — that an out-of-range pair raises before any slice is taken — which is a property of the check and needs no fault to observe. `test_the_predicate_refuses_directly` runs first and drives `_require_readable`, which decides without touching memory, so if the guard is ever removed from the readers that cell fails cleanly and names the cause before any cell could reach a raw slice. An earlier revision of this branch used subprocesses to observe the pre-fix SIGBUS; that produced host crash reports on every run and has been removed. The pre-fix behaviour is recorded once, in #1442.

## Spec

§9.5.6 now states the validation rule, which is where §13.7 already pointed for it.

Fixes #1442

## Summary by CodeRabbit

- **New Features**
  - `vera serve` and the WASI HTTP server now validate handlers against the declared Vera `Request` and `Response` types.
  - Diagnostics identify invalid, shadowed, or missing handler type information.

- **Bug Fixes**
  - Invalid memory reads are safely rejected with execution errors.
  - Bounds checks now apply consistently across server, heap, and Markdown data readers.

- **Documentation**
  - Updated the specification and project documentation with handler validation rules and revised test-suite metrics.


aallan (commented):
Reviewed at `a6ebe593` on base `bcaae63f`, fresh worktree, canary before every run. I first reviewed `8e8d1ad9`; the rebase moved the base but the PR's own content is unchanged — `vera/codegen/api.py`, `vera/runtime/heap.py`, `vera/runtime/server.py`, `tests/test_handler_signature_1442.py` and `spec/09-standard-library.md` are byte-identical across it, and `vera/codegen/core.py` differs only by #1419's per-owner constructor hunks. Every measurement below is against `a6ebe593`.

**On the host-crash rule.** No probe of mine faulted. Everything ran through a wrapper that installs `SIGBUS`/`SIGSEGV`/`SIGILL`/`SIGFPE` handlers calling `os._exit(200)` under a `perl -e 'alarm 25'` cap, and — more importantly — every probe was designed to stop *at the guard*, so no raw slice was ever reached. All reported `PROBE_EXIT=0`; none reported 200. I deliberately did **not** run the out-of-range readers against the base, did not pass `--crash-host`, and did not send a request to the base's `vera serve` when it accepted the bad handler.

**Audit of the PR's own tests: clean.** `tests/test_handler_signature_1442.py` contains no `subprocess` call at all, and its docstring says in as many words that nothing there reproduces the crash — the cells assert the guard raises before any slice. The only repo test mentioning `make_server` is `tests/test_cli.py`, in a comment, and it is untouched.

## The fix works

Red-first, same nine programs on both sides:

| program | base `bcaae63f` | head `a6ebe593` |
|---|---|---|
| **the issue's own** `keep(@Request -> @Response)` + `handle(@Bool -> @Bool)` | **ACCEPTED** | refused — "its parameter is not a data type, so it cannot be @Request; it returns not a data type…" |
| `handle(@Request -> @Bool)` | **ACCEPTED** | refused — names the return |
| `handle(@Request -> @Widget)` | **ACCEPTED** | refused — "it returns @Widget, not the prelude @Response" |
| `handle(@Widget -> @Widget)` (unrelated ADT, same i32 ABI) | **ACCEPTED** | refused — both positions named |
| entry declares `data Request` (shadows prelude) | **ACCEPTED** | refused — "this program's own @Request, which shadows the prelude's" |
| entry declares `data Response` | **ACCEPTED** | refused — the mirror message |
| entry declares **both** | **ACCEPTED** | refused — both clauses |
| `type Req = Request` onto a *shadowed* `Request` | **ACCEPTED** | refused — the alias resolves and the shadowing is still caught |
| `handle(@Request -> @Future<Response>)` | **ACCEPTED** | refused |

Controls unchanged: `@String -> @Response` and two-parameter handlers were already refused; the valid handler and a `type Req = Request` / `type Resp = Response` alias pair are accepted on both.

Further shapes, all correct at head: `Option<Response>` return → "@Option, not the prelude @Response"; `Option<Request>` parameter → likewise; `Box<Int>` return → "@Box" (the documented drop-the-type-arguments rule); a refinement `{ @Request | true }` → accepted; a two-level alias chain → accepted; a `private handle` → refused for absence; a module-declared `data Request` → blocked upstream at **E621** on both revisions; a module-declared `handle` → "must export a public 'handle'" on both.

**Both surfaces, one guard.** `vera compile --target wasi-p2 --world server` refuses the same set through the shared `validate_handler` (`vera/codegen/wasi.py:4851`). At base, `bad_bool` and `bad_ret` compiled to a server component, and the shadowed-`Request` program died with a raw `KeyError: 'Request'` traceback — head turns all three into the structured diagnostic. `vera serve` likewise refuses before listening, where base started the accept loop on the `Bool` handler.

**The readers.** 17 out-of-range pairs — offsets at, past and straddling the limit, negative offsets, negative lengths, lengths overflowing the offset, `2**40` and `2**62` — every one refused with the `out of bounds memory access` reason `execute()` classifies, and the in-range controls (including exactly-at-the-limit and a zero-length read at the limit) still work.

**The external reproduce script** at this head aborts at F2 with the diagnostic instead of printing "INVALID … HANDLER ACCEPTED". F1/F3/F4 are other findings and sit outside this diff.

**`vera serve` end to end** on `examples/http_server.vera`: `GET /` 200/15 bytes, `POST /echo` 200/5 bytes, `GET /nope` 404.

## Findings

| # | Sev | What |
|---|-----|------|
| 1 | **MED** | **The CHANGELOG's containment claim is false as written.** It says "`_read_i32_at` and `_read_bytes_at` were the last two raw readers with no bounds check". Two more remain at this head: `_read_i32` (`vera/runtime/heap.py:1054`) and `_read_f64` (`:1064`), both module-level, both slicing `memory.data_ptr(caller)` raw, neither calling `_require_readable`. They are not dead: `vera/runtime/json.py` imports both and passes them into `read_json` on the `json_stringify` host path (`json.py:134`), and `json.py` contains no bounds check of any kind. Not a regression — base has all four unguarded — and not an active vulnerability, since no validator hole feeds them a bad pointer today. But the sentence is measurably wrong, and the containment stops one decoder short of the family it names, which is the family the same paragraph promises to have closed. Related: `heap.py` now carries **three** copies of one predicate — `_require_readable` (`:134`), `_read_wasm_string`'s inline check (`:72`), and the DB decoder's local `_require_in_bounds` (`:1203`) — the keep-in-step hazard the PR's own docstring is about. |
| 2 | **MED** | **A surviving mutation: alias resolution is untested.** Deleting the `NameSort.ALIAS` / `ALIAS_ARITY_MISMATCH` branch from `_declared_adt_name` leaves the whole file **green** (17 passed with the reader class deselected) — while `type Req = Request; type Resp = Response; public fn handle(@Req -> @Resp)` flips from ACCEPTED to **refused**, a false refusal of a legitimate program with the message "its parameter is not a data type". The docstring states the property explicitly ("so a `type Req = Request` alias resolves"), and no fixture exercises it. Against the "12 mutations / 0 survivors" claim, this is a survivor; one alias fixture beside `VALID` closes it. (My other three mutations were killed: dropping the provenance half of `_is_prelude_adt` reds the two ownership cells; disabling the return check reds `right parameter, wrong return`; neutering `_require_readable` reds `test_the_predicate_refuses_directly`.) |
| 3 | LOW | **The reader class's ordering is legibility, not containment.** `test_the_predicate_refuses_directly` does fail cleanly when the guard is neutered — I measured exactly that, and it cannot fault because it drives the predicate directly. But `pyproject.toml`'s `addopts = "-m 'not stress'"` carries no `-x`, so pytest continues, and the very next cell (`test_read_i32_at_is_guarded`) would reach a raw out-of-range slice and take the worker down with the fault the class exists to avoid. The docstring reads as though the ordering prevents that ("fails cleanly and names the cause *before* any cell that would reach a raw slice"). A precondition assert, or a skip driven by the predicate's own verdict, would make it true. |
| 4 | INFO | `_declared_adt_name`'s docstring says `Future<T>` "is not `T`; a handler returning `Future<Response>` returns a future", which reads as though the refusal would name `@Future`. Measured: it answers `None`, so the message is "it returns not a data type". Right outcome, different route than documented. |
| 5 | INFO | #1442 is an open `bug` with no row in `KNOWN_ISSUES.md`'s Bugs table or the v0.2.0 burndown at **either** revision, so nothing was removed here and both tables stay self-consistent by never having accounted for it. Pre-existing parity gap, not this PR's to fix — noted only because the fix closes a bug the queue never listed. |
| 6 | INFO | `test_refused` matches on `"Request"`, which appears in the fixed prefix of every refusal, so that cell alone cannot tell one refusal from another. `TestTheRefusalNamesTheRemedy` supplies the discrimination, so the pair is covered — worth knowing if the parametrisation is ever read on its own. |

## Verified green

- **Corpus differential vs `bcaae63f`: 0 movers** — 303 programs, 254 identical WAT, 49 compiled at neither. C3's figure reproduced.
- `pytest tests/` **13,380 passed / 191 skipped / 26 deselected**, matching TESTING.md's row exactly · `mypy vera/` clean (106 files) · `ruff check .` and `ruff check --select S vera/` clean.
- 16 gate scripts green: 253 conformance, 43 examples, `check_examples_run`, 303 corpus canonical, `check_doc_counts` (**13,597 tests, 204 files, 253 conformance, 43 examples**), site assets, diagnostic fields, limitations sync, explicit encoding, version sync, and the spec/SKILL/README/FAQ example gates. `VERA_JS_COVERAGE=1 pytest tests/test_browser.py` **425 passed**.
- Docs: the rule lands in **spec §9.5.6 HttpServer**, which is exactly the section `spec/13-wasi.md:174` already cites for the server world, so §13 needed no edit; §12 is the browser target, where `<HttpServer>` is unavailable; `LSP_SERVER.md` does not mention handler validation. TESTING.md gains the new file's row and the counts are re-derived.

The guard itself is right, and I could not find a signature that gets past it. Finding 1 is a sentence to correct (and, if you want the claim to be true, two readers to guard); finding 2 is one fixture.

VERDICT: OPEN — 2 findings (a false CHANGELOG claim about the containment's completeness, and a surviving mutation showing alias resolution is untested), 4 informational; the fix for #1442 itself is correct and complete

aallan:
@coderabbitai Confirmed and fixed in 7f165fc5: `markdown._read_string` now goes through `_require_readable` before `_slice_and_decode`, with a cell covering the past-the-end, straddling, negative and overflowing pairs and an in-range decode as the control. The family enumeration in the test, the `_require_readable` docstring and the CHANGELOG entry now name it, and the enumeration rule reads "every raw slice of `memory.data_ptr`, direct or through a shared helper", which is how this one was missed.

